PR #22, opened 2020-01-17
valine comment integration

caryyu:
Refers to: https://valine.js.org/en/jekyll.html


caryyu:
Don't worry about this AK as it is mentioned in the official doc, which at least doesn't belong to me

cotes2020:
Sorry, I don't need another comment system.

PR #25, opened 2020-02-16
Run and build inside a Docker container

Blafy:
To avoid installing all the dependencies (python + ruby + deps) in order to build the website files, I added a Dockerfile.

1. To build the Docker image, run `Docker build -t chirpy-build .`

2. To run locally, just type `Docker run --rm -v ${PWD}:/app chirpy-build tools/run.sh`

3. To build locally, type `Docker run --rm -v ${PWD}:/app chirpy-build tools/build.sh`. The resulting files are in the _site folder

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,4 +16,4 @@ RUN bundle install
 
 CMD ["/bin/sh"]
 
-ENTRYPOINT ["./tools/run.sh", "-h"]
+ENTRYPOINT ["./tools/run.sh", ""]
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,5 +15,3 @@ RUN gem install jekyll -v3.8
 RUN bundle install
 
 CMD ["/bin/sh"]
-
-ENTRYPOINT ["./tools/run.sh", ""]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+FROM ruby:2.4-slim
+
+WORKDIR /app
+
+COPY Gemfile .
+
+RUN apt-get -y update && apt-get -y install python3 python3-pip make gcc  g++ git
+
+RUN pip3 install ruamel.yaml
+
+RUN ln -s /usr/bin/python3 /usr/bin/python
+
+RUN gem install jekyll -v3.8
+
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM ruby:2.4-slim
```
cotes2020: ruby `2.4` is too old, `2.6.5-slim` might be a better choice

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+COPY Gemfile .
+
+RUN apt-get -y update && apt-get -y install python3 python3-pip make gcc  g++ git
+
+RUN pip3 install ruamel.yaml
+
+RUN ln -s /usr/bin/python3 /usr/bin/python
+
+RUN gem install jekyll -v3.8
```
cotes2020: This step should is unused, it should be removed.

cotes2020:
```suggestion
RUN gem install jekyll -v3.8
```
This step is unsed.

Blafy: I used this line to fix the version of Jekyll, to avoid all updates and breaking changes of Jekyll 4+. Unless your code is compatible with Jekyll 4.0 ?

cotes2020: Check this: <https://github.com/cotes2020/jekyll-theme-chirpy/blob/cf6c94a3845a0c4911e2eaf6490699ce31988a86/Gemfile#L3>

Gemfile already indicated the Jekyll version, so the line `RUN gem install jekyll -v3.8` won't work.

Blafy: Oh, you're right.
I thought your Gemfile was installing 4.0, my memory is a bit blurry (I built this image a month ago https://hub.docker.com/r/blafy/chirpy-build).
I will fix the dockerfile then :)

cotes2020: One more thing, I build the image locally (on mac) and found the size  up to **834 MB**, could you reduce the size?

```console
$ docker image ls
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
chirpy-pr           latest              582049aa1b5f        19 hours ago        834MB
```

PR #51, opened 2020-05-03
Fix to search result overflow

karlsandin:
Adds word break on search result page preamble, this bug is caused when using code markdown early in a post a can be seen in the screenshot below.

image

### DIFF
--- a/assets/css/main.scss
+++ b/assets/css/main.scss
@@ -384,6 +384,7 @@ $sidebar-display: "sidebar-display";
 }
 
 #search-results {
+  word-break: break-word;
   padding-bottom: 6rem;
   a {
     @extend %link-color;
--- a/assets/css/main.scss
+++ b/assets/css/main.scss
@@ -384,7 +384,6 @@ $sidebar-display: "sidebar-display";
 }
 
 #search-results {
-  word-break: break-word;
   padding-bottom: 6rem;
   a {
     @extend %link-color;
@@ -1074,6 +1073,10 @@ table {
     .fa-times-circle {
       right: 5.2rem;
     }
+  }
+
+  #search-results>div {
+    max-width: 100% !important;
   }
 
   #search-input {


cotes2020:
Sorry, I'm not sure what is meant by “when using code markdown early in the post”, could you please provide more details on how to reproduce the bug? (Maybe just paste a code snippet down below)

karlsandin:
Sure, sorry for being unclear! :)

If I create a post and add this code snippet as the first thing in the post the search result preamble/description will then add the snippet row number as one long word and therefore break the design.

Hope this made it a bit clearer!

```jsx
// EmailInput wraps an HTML `input` and adds some app-specific styling.
const EmailInput = React.forwardRef((props, ref) => (
  <input ref={ref} {...props} type="email" className="AppEmailInput"></input>
));

class App extends Component {
  emailRef = React.createRef();

  // `this.emailRef.current` points to the `input` component inside of EmailInput,
  // because EmailInput is forwarding its ref via the `React.forwardRef` callback.
  onClickButton() {
    this.emailRef.current.focus();
  }

  render() {
    return (
      <div>
        <EmailInput ref={this.emailRef} />
        <button onClick={() => this.onClickButton()}>
          Click me to focus email
        </button>
      </div>
    );
  }
}
```

assets/css/main.scss
```diff
@@ -384,6 +384,7 @@ $sidebar-display: "sidebar-display";
 }
 
 #search-results {
+  word-break: break-word;
```
cotes2020, line 387: `break-word` will reduce the reading experience, don't use it.

Add `max-with: 100%;` between line 399 to 400 is the better approche

karlsandin: Good point! 👍

cotes2020:
I have combined your two commits into one (see e8630ab556d8fcc6f09a7982d3836af1a1c89261), thanks for the PR.

Issue #41, opened 2020-03-22
Posts discovery: Allow finding posts in subdirectories of _posts folder


ppipada:
**Is your feature request related to a problem? Please describe.**

When writing blog posts, I find it useful to arrange posts file (*.md) in appropriate subfolders. This helps in keeping posts and associated files together for easy maintenance.  
E.g: If I am writing a post about python and I have a few python files which I would like to include as code snippets via liquid syntax, I find it maintainable to keep all such python posts and files in subdirectories as `_posts/python/`. 

If we do this currently; pages, categories and tags of `_posts` folder are populated but not from subfolders.   

**Describe the solution you'd like**

Pages, categories and tags of `_posts` folder and any subdirectories are populated.

**Describe alternatives you've considered**

We can run the build python scripts (init/run/build, etc) per directory as explicit input, but doing this for multiple such directories becomes problematic. 

**Possible solution**
- In file update_posts_lastmod.py, line 130
```
<<< path = os.path.join(POSTS_PATH, "*.md")
>>> path = os.path.join(POSTS_PATH, "**", "*.md")
```

- In file pages_generator.py, line 67 and 132.
```
<<<         for file in glob.glob(os.path.join(path, '*.md')):
>>>         for file in glob.glob(os.path.join(path, '**', '*.md')):

```
I have tried and tested the solution at my [personal site](https://ppipada.github.io/) at [github repo](https://github.com/ppipada/ppipada.github.io)

cotes2020:
Um... put posts together and sort by date is enough; the blogger don't need to spend time creating subfolders.

r0bsec:
@ppipada where exactly did you make this change? It looks like you are referencing a Python file, but there is no Python in this project.

How can I modify an existing site?

---

@cotes2020 I'd like to support subfolders so that I can have a folder per blog post/article. If one post is made of up an `/project1/<date>_<title>.md` then in that `/project1/` folder I might have related images, screenshots, logs, etc.

Without folder support, the one `_posts` folder becomes a junk drawer of many, many files and it's impossible to tell which post `screenshot2.png` is related to.

**Please re-consider taking this on as a feature request!!**

ppipada:
@r0bsec this was in 2020. Version 2.x. 
Not sure how this is handled today. I have moved to Hugo for my blog, so no current context.

PR #87, opened 2020-06-29
_scripts: Take care of posts in subdirectories

vj-kumar:
We could create subdirectories inside _post directory and add
posts inside those. Take care of those posts while creating
categories and tags.

Signed-off-by: Vijai Kumar K <vijaikumar.kanagarajan@gmail.com>


cotes2020:
It's duplicate #41

Issue #42, opened 2020-03-22
Post excerpts: Use post.excerpt syntax for home page content


ppipada:
**Is your feature request related to a problem? Please describe.**

For posts that have < 200 chars first paragraphs, or posts with headings and bulleted points, or posts with headings and code snippets, current per-post content (i.e stripped excerpts) on home page get incomprehensible. 

This is mainly because the current post content is take as `{{ post.content | strip_html | truncate: 200 | replace: '&', '&amp;' }}`  in `_layouts/home.html`.

**Describe the solution you'd like**

Use Jekyll` {{ post.excerpt }}` syntax for getting the post summary: https://jekyllrb.com/docs/posts/#post-excerpts. 

This would also allow use of `excerpt_separator ` in `_config.yml` as needed. 

**Solution**
In `_layouts/home.html` line 18
```
<<< {{ post.content | strip_html | truncate: 200 | replace: '&', '&amp;' }}
>>> {{ post.excerpt }}
``` 

I have tried and tested the solution at my [personal site](https://ppipada.github.io/) at [github repo](https://github.com/ppipada/ppipada.github.io)

cotes2020:
Jekyll tag `excerpt` is good, but, truncate the 200 words of posts could make the layout of homepage more neat.

ppipada:
Truncate the 200 words of posts definitely becomes neat for posts that are paragraphs of text. For posts that are brief heading and then bullet list or numerical numbered lists it absolutely messes up.

jexxers:
@cotes2020 - in what situations would `post.content` look more tidy than `post.excerpt`?

I was about to submit a PR, but saw others have also encountered this issue.

joce:
Could we make it optional? If your post doesn't have a defined `excerpt_separator`, truncate to 200 words, otherwise, use `excerpt`?

abd3lraouf:
Created a PR with a solution to this #619

Seblor:
Any update on that ? I saw that #619 was closed, but It's still an issue.

abd3lraouf:
@Seblor please feel free to take what I did and add it here :-) 💯

bobrocke:
@cotes2020 I know I am very late to this discussion, but I just started working with Chirpy and find the lack of excerpt functionality a problem. Yes, poor use of such an option could make the home page messy, but that's sort of up to the Blog writer and not so much the theme.

Would you reconsider the PR?

cotes2020:
[...]

PR #143, opened 2020-10-04
Changed post list rendering to use post.excerpt

amkherad:
## Changed post list rendering to use post.excerpt instead of post.content to respect jekyll's excerpt, so user would set whatever he wants as post excerpt.

only _layouts/home.html has changed.

## Type of change

- [ X] Bug fix (It changes how post lists are shown)

## How has this been tested

- [ X] I have tested this feature in the browser

### Test Configuration

- Browerser type & version:
- Operating system: ArchLinux
- Bundler version: Bundler version 2.1.4
- Ruby version: ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x86_64-linux]
- Jekyll version: jekyll 4.1.1

### DIFF
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -58,7 +58,7 @@ layout: page
     </div>
     <div class="post-content">
       <p>
-        {% include no-linenos.html content=post.content %}
+        {% include no-linenos.html content=post.excerpt %}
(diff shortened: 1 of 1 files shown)


cotes2020:
It's duplicate to #42

PR #250, opened 2021-01-15
added text-muted class to typography to override BS

RyanTG:
## Description

I noticed that bootstrap has `!important` on their text-muted class, which made it so that we couldn't control that color. So I added that class to the light/dark typography files, with an !important. And now we can adjust that color!

Let me know if you like this, but want it adjusted (like where I put the class in those files).

e.g. Fixes #(issue)

CSS update.

## How has this been tested

### DIFF
--- a/assets/css/_colors/dark-typography.scss
+++ b/assets/css/_colors/dark-typography.scss
@@ -136,4 +136,8 @@
       rgb(26, 26, 30));
   }
 
+  .text-muted {
+    color: var(--text-muted-color) !important;
+  }
+
 } // dark-scheme
--- a/assets/css/_colors/light-typography.scss
+++ b/assets/css/_colors/light-typography.scss
@@ -77,4 +77,8 @@
   --footer-bg-color: #ffffff;
   --footnote-target-bg: lightcyan;
   --footer-link: #424242;
+
+  .text-muted {
+    color: var(--text-muted-color) !important;
+  }
 } // light-scheme


cotes2020:
I don't think it is necessary to change the class `.text-muted` of Bootstrap. 

If you are not satisfied with its color, you can define your own styles to the specified CSS selector instead of overwriting it.

cotes2020:
> you might want to override all instances ...

In fact, I don't want to cover any of them.

RyanTG:
Ha, I didn't actually mean *you* :D Just anyone who doesn't like the grey from bootstrap. 

My usage might be more unique, because I am changing every color in the theme.

On that note: I noticed that the search placeholder text color can only be changed for the ::focused selector, but not for non-focused. Are you ok with that? I was considering submitting a PR. But if you don't want that, then I won't.

Thanks again for the nice theme!

cotes2020:
The purpose of the design is that when the search input field is focused, the placeholder text becomes lighter and highlights the text being typed.

I really appreciate your enthusiasm for reading the source code, but as the [contributing guidelines][instruction] says,  it is best not to modify the theme color in PR, because everyone has different preferences for colors. If I merge all of them into one theme, the colors will be confused. Hope you can understand.

BTW, if it is a PR about CSS code optimization, I'd welcome it.

[instruction]: https://github.com/cotes2020/jekyll-theme-chirpy/blob/master/.github/CONTRIBUTING.md

RyanTG:
Thanks. I certainly would not modify theme colors in a PR. I was only trying to give users the *option* of modifying some that currently are difficult to modify.

PR #297, opened 2021-03-14
Fix SW Error: Request method 'POST' is unsupported

xcaro:
## Description

Fix issue related to Service Worker

image

e.g. Fixes #(issue)

## Type of change

- [ ] Documentation update

## How has this been tested
- [x] N/A
## #Test Configuration
- [x] N/A
### Checklist
- [x] N/A

### DIFF
--- a/assets/js/data/cache-list.js
+++ b/assets/js/data/cache-list.js
@@ -60,6 +60,7 @@ const exclude = [
   {%- if site.google_analytics.pv.proxy_url and site.google_analytics.pv.enabled -%}
     '{{ site.google_analytics.pv.proxy_url }}',
   {%- endif -%}
+  'https://www.google-analytics.com',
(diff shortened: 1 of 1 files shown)


cotes2020 (commented):
Thanks for the PR!

I think the reason for this error is that `sw.js` (service worker) stores the `Request` with the POST method as the key, which is not supported by the Web API.

Simply removing the POST request url (starting with the Google Analytics domain) did not completely solve this problem, so I think that improving the caching strategy of `sw.js` is a better way.

xcaro:
Hi @cotes2020 ,

I'm not specialist on it. Can you suggest some idea to improve the caching?

Thanks,

cotes2020:
My solution is 2f2832c

xcaro:
This is looking really good solution.

Thanks you @cotes2020 ,

PR #420, opened 2021-10-03
Prevent unwanted line break (Safari)

MartinP7r:
## Description

On Safari the first `div` of the flex-box gets wrapped at the date's day.  
My suggestion would be to add `text-nowrap` to either the flex-box itself or just the first `div` (below "Published date and author").

Please feel free to make relevant changes or let me know if this is not a change you would want to make.

| Safari | Chrome |
| ----- | ----- |
|Screen Shot 2021-10-03 at 20 09 12 | Screen Shot 2021-10-03 at 20 09 25|

## Type of change

- [ ] Documentation update

## How has this been tested

**This is not possible. Error:** `ERROR: This script is not allowed to run outside of GitHub Action.`

### Test Configuration

- Browser type & version: Safari, Chrome (latest)
- Operating system: Mac OSX Big Sur
- Bundler version: 1.17.2
- Ruby version: 3.0.3
- Jekyll version: 4.2.1

### DIFF
--- a/_layouts/post.html
+++ b/_layouts/post.html
@@ -13,7 +13,7 @@ layout: default
 
       <h1 data-toc-skip>{{ page.title }}</h1>
 
-      <div class="post-meta text-muted d-flex flex-column">
+      <div class="post-meta text-muted text-nowrap d-flex flex-column">
         <!-- Published date and author -->
         <div>
           <span class="semi-bold">


cotes2020:
First of all, thanks for the PR. However, when I use Safari (macOS) to visit the demo site, I don’t see the wrap problem

Screen Shot 2021-10-03 at 20 29 42

If you can be more specific about how to reproduce this issue, that would be great

cotes2020:
> What I noticed is that your demo version and my published version have quite differently formatted source code. E.g. the demo has a "denser" head tag.

It seems that Safari reserves space for HTML comments. As for how to have compressed output like the demo site, please finish reading the tutorial.

MartinP7r:
Thank you for your time. I'm still not sure what causes the difference but will investigate further. I'm pretty sure the problem is with my setup.  
I've read through https://chirpy.cotes.info but I don't think I've seen any mention of compression. Is there any other tutorial?  
My guess is it's the `compress_html` setting in `_config.yml`, but mine is set up exactly like the source project.

PR #453, opened 2021-11-30
Accessibility Bug Fix: Tab focus doesn't reach Dark mode toggle button

bhavykhatri:
## Description
This is a fix for accessibility bug

- if you press tab then the focus doesn't reach dark mode toggle button
- After focus should reach the dark mode toggle button, pressing enter button should change the theme of website
image

e.g. Fixes #(issue)
-->

## Type of change

- [ ] Documentation update

## How has this been tested

### Test Configuration

- Browser type & version:
- Operating system: Windows
- Ruby version: 2.7.4p191 
- Bundler version:  2.2.30
- Jekyll version: 4.2.1

### Checklist

### DIFF
--- a/_includes/mode-toggle.html
+++ b/_includes/mode-toggle.html
@@ -147,6 +147,6 @@
     {
       toggle.flipMode();
     }
-});
+  });
 
 </script>
--- a/_includes/mode-toggle.html
+++ b/_includes/mode-toggle.html
@@ -143,8 +143,7 @@
   });
 
   $("#mode-toggle-wrapper").keyup(function(e){
-    if(e.keyCode == 13)
-    {
+    if(e.keyCode == 13) {
       toggle.flipMode();
     }
   });
--- a/_includes/mode-toggle.html
+++ b/_includes/mode-toggle.html
@@ -142,4 +142,11 @@
 
   });
 
+  $("#mode-toggle-wrapper").keyup(function(e){
+    if(e.keyCode == 13)
+    {
+      toggle.flipMode();
+    }
+});
+
 </script>
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -52,6 +52,15 @@
   </ul> <!-- ul.nav.flex-column -->
 
   <div class="sidebar-bottom mt-auto d-flex flex-wrap justify-content-center align-items-center">
+    {% unless site.theme_mode %}
+      {% if site.data.contact.size > 0 %}
+        <span class="icon-border order-2"></span>
+      {% endif %}
+
+      <span id="mode-toggle-wrapper" class="order-1" tabindex="0">
+        {% include mode-toggle.html %}
+      </span>
+    {% endunless %}
 
     {% for entry in site.data.contact %}
       {% capture url %}
@@ -78,17 +87,6 @@
       {% endif %}
 
     {% endfor %}
-
-    {% unless site.theme_mode %}
-      {% if site.data.contact.size > 0 %}
-        <span class="icon-border order-2"></span>
-      {% endif %}
-
-      <span id="mode-toggle-wrapper" class="order-1">
-        {% include mode-toggle.html %}
-      </span>
-    {% endunless %}
-
   </div> <!-- .sidebar-bottom -->
 
 </div><!-- #sidebar -->
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -52,6 +52,7 @@
   </ul> <!-- ul.nav.flex-column -->
 
   <div class="sidebar-bottom mt-auto d-flex flex-wrap justify-content-center align-items-center">
+    
     {% for entry in site.data.contact %}
       {% capture url %}
         {%- if entry.type == 'github' -%}
@@ -87,7 +88,7 @@
         {% include mode-toggle.html %}
       </span>
     {% endunless %}
-    
+
   </div> <!-- .sidebar-bottom -->
 
 </div><!-- #sidebar -->
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -52,16 +52,6 @@
   </ul> <!-- ul.nav.flex-column -->
 
   <div class="sidebar-bottom mt-auto d-flex flex-wrap justify-content-center align-items-center">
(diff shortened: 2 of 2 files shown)


bhavykhatri:
@cotes2020

_includes/mode-toggle.html
```diff
@@ -142,4 +142,11 @@
 
   });
 
+  $("#mode-toggle-wrapper").keyup(function(e){
+    if(e.keyCode == 13)
+    {
+      toggle.flipMode();
+    }
+});
```
bhavykhatri, line 150: Fix indentation

_includes/sidebar.html
```diff
@@ -52,6 +52,15 @@
...
 
   <div class="sidebar-bottom mt-auto d-flex flex-wrap justify-content-center align-items-center">
+    {% unless site.theme_mode %}
+      {% if site.data.contact.size > 0 %}
+        <span class="icon-border order-2"></span>
+      {% endif %}
+
+      <span id="mode-toggle-wrapper" class="order-1" tabindex="0">
+        {% include mode-toggle.html %}
+      </span>
```
cotes2020, line 62: Please restore the changes of `_includes/sidebar.html`

_includes/mode-toggle.html
```diff
@@ -142,4 +142,11 @@
 
   });
 
+  $("#mode-toggle-wrapper").keyup(function(e){
+    if(e.keyCode == 13)
+    {
```
cotes2020, line 147: In order to maintain a consistent code style, please modify it as follows:

```js
$("#mode-toggle-wrapper").keyup(function(e) {
    if(e.keyCode == 13) {
    // ...
```

cotes2020 (changes requested):

_includes/mode-toggle.html
bhavykhatri: Yes you are right, will do.

_includes/sidebar.html
cotes2020: > tab order for darkmode should come before other social media link...

In the design expectation, this is to make the mode-switching JS code not block the rendering of social icons, which is why `.order-x` is specified.

> The best option would be to use button which would save us from the hassle of using custom event listeners...

Can you please explain why the button is better than EventListener in this case?

bhavykhatri: - How about if instead of inlining js we use <script async/defer src="mode-toggle"> javascript. In this way rendering of social icons won't be blocked.
- Note that I have added one attribute tabindex="0", to have the correct tab order, in case you missed.
- checkout this for button soln, https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_button_test, although only onclick was specified but when you tab focus it and press enter it still works.

cotes2020: Sorry, I missed the `tabindex=0`, you can keep it, but please restore the code block to the bottom. My goal is to execute the mode detection JS immediately. If we use `async/defer`, when the session has a theme-mode state and the user jumps to another page, there will be a dark/light switch flicker, so please be sure to keep the original JS running priority.

In addition, the button is just another method of implementation, and there is no outstanding advantage to replacing the event listener.

bhavykhatri: Agreed with your first point.

Regarding advantage of using button element, for the people in accessibility need when they use screen reader, it explicitly says that this is a "button" which is not the cases with using span element. Also, for the programmer it is more intuitive when he/she is going through the code. Yes at the end result will be same visually but again not all people consume/interact with the web content visually. I hope you get my point. Although not making any further changes for now. Thanks and have a nice day :)

_includes/sidebar.html
```diff
@@ -84,7 +84,7 @@
         <span class="icon-border order-2"></span>
       {% endif %}
 
-      <span id="mode-toggle-wrapper" class="order-1">
+      <span id="mode-toggle-wrapper" class="order-1" tabindex="0">
```
cotes2020, line 87: 😂 Sorry again, I missed the `tabindex` of other icons, maybe you need to make other changes:

1. Line 74 of [`_includes/sidebar.html`](https://github.com/bhavykhatri/jekyll-theme-chirpy/blob/e73e3b26801830b64bb16c6547a58afdf23f76b1/_includes/sidebar.html#L74) should be changed as follow:

```diff
- {% unless site.theme_mode %}class="order-{{ forloop.index | plus: 2 }}"{% endunless %}
+ {% unless site.theme_mode %}class="order-{{ forloop.index | plus: 2 }}" tabindex="{{ forloop.index | plus: 2 }}"{% endunless %}
```

2.  Maybe `#mode-toggle-wrapper`  would be better to use `tabindex="1"`, so we can get index from `1` to `n`

```diff
- <span id="mode-toggle-wrapper" class="order-1" tabindex="0">
+ <span id="mode-toggle-wrapper" class="order-1" tabindex="1">
```

cotes2020 (changes requested):

_includes/sidebar.html
bhavykhatri: For both of the points, anchor tag are already tab focusable therefore no need for the changes. Tabindex are only used in special elements where the element like span is not tab focusable but to make it tabfocusable.

bhavykhatri: https://jsfiddle.net/rq7j1ea8/

Checkout this js fiddle to understand more

have a look at this documentation if you have time https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

cotes2020: Thank you for your reminder, I have actually read this article(<https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex>) before

I tested your PR code. If we don’t add `tabindex` to other icons, we will get the weird focus order as mentioned in the attached image above:

tabindex

_includes/sidebar.html
cotes2020: > for the programmer it is more intuitive when he/she is going through the code...

True, I agree with this 👍

_includes/sidebar.html
cotes2020: NVM, it may be the best to automatically determine the tab order of other elements, so this PR is merged. Thank you for your time！

PR #469, opened 2021-12-30
Bug Fix: Link in a heading is rendered as an anchor and becomes invisible.

iwasaki501:
## Description

`<a>` tags in headings are set to apply the style of `%anchor` in `module.scss` and `common.scss`. Therefore, if I put a link in a heading, the link will have the same style as the anchor, which means that the font size becomes smaller and the link will not be displayed unless I hover the mouse over it.

For example, 
```
## Header with a [link](https://example.com)
```
in a markdown file is displayed like this.

Animation

My suggestion is to set an `anchor` class for the `<a>` tag of the anchor and apply style settings to the `anchor` class, not to the `a` tag.

## Type of change

- [ ] Documentation update

## How has this been tested

### Test Configuration

- Browser type & version: Google Chrome 96.0.4664.110
- Operating system: Windows 10 21H1 + WSL (18.04.1 LTS)
- Ruby version: 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux-gnu]
- Bundler version:  2.2.15
- Jekyll version: 4.2.1

### Checklist

### DIFF
--- a/_includes/refactor-content.html
+++ b/_includes/refactor-content.html
@@ -169,7 +169,7 @@
       {% endif %}
 
       {% assign id = snippet | split: '"' | first %}
-      {% capture anchor %}<a href="#{{ id }}"><i class="fas fa-hashtag"></i></a>{% endcapture %}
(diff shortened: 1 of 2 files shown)


cotes2020 (approved):
Thanks for the patch 👍

PR #478, opened 2022-01-04
update english locales for recently updated posts

joshjohanning:
## Description

This is a very minor wording fix/suggestion, but I think that it is more grammatically correct to say `Recently Updated` vs. `Recent Update`. `Recent Update`, to me, implies singularity whereas `Recently Updated` could imply single or multiple updated post(s).

I think `Recently Updates` could also work here, depending on preference, since it matches `Tags` and `Table of Contents`.  But the problem would be if there are only 1 recently updated post, it would be dramatically incorrect. 

Really splitting hairs here, but I made the change on my blog so thought I would contribute back. 

Feel free to reject if you disagree! 🙂

## Type of change

## How has this been tested

- [ ] I have tested this feature in the browser

### Test Configuration

- Browser type & version: `Brave (chromium) 1.33.106`
- Operating system: `macOS 11.6.3`
- Ruby version: `ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-linux]`
- Bundler version: `Bundler version 2.2.26`
- Jekyll version:   `* jekyll (4.2.1)`

### Checklist

### DIFF
--- a/_data/locales/en.yml
+++ b/_data/locales/en.yml
@@ -23,7 +23,7 @@ search:
   no_results: Oops! No result founds.
 
 panel:
-  lastmod: Recent Update
+  lastmod: Recently Updated
   trending_tags: Trending Tags
   toc: Contents
 


cotes2020 (approved):
Thanks for sharing 👍

PR #482, opened 2022-01-14
Don't include Google Analytics scripts without analytics id

Tamschi:
## Description

I'm in the EU, so I have to turn off Google Analytics entirely to avoid legal issues. I noticed that the Google Analytics script was still being loaded even though I hadn't changed `google_analytics.id` from the empty string in `_config.yml`.

I'm not sure if the first part of the conditions is still needed, since I don't know the Liquid templating language.
I suspect it's necessary in case someone comments out that part of the configuration.

This is a bit of a "maybe you can use this" PR, since I'm maintaining my fork separately just for my page (because I customised a number of things outside the configuration files) and won't necessarily upgrade to the next version of Chirpy.

## Type of change

## How has this been tested

  (I'm on Windows, which is probably (part of) why this doesn't work for me. I did use bash.)

### Test Configuration

- Browser type & version: Brave Version 1.34.80 Chromium: 97.0.4692.71 (Official Build) (64-bit)
- Operating system: Windows 10
- Ruby version: ruby 3.0.3p157 (2021-11-24 revision 3fb7d2cadc) [x64-mingw32]
- Bundler version: Bundler version 2.3.4
- Jekyll version: * jekyll (4.2.1)

### Checklist

  (It's a two-line adjustment that should be self-explanatory.)
  (Or rather: There was nothing to update, since this doesn't change features.)

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -40,7 +40,7 @@
   <link rel="dns-prefetch" href="https://fonts.gstatic.com">
 
   <!-- GA -->
-  {% if jekyll.environment == 'production' %}
+  {% if jekyll.environment == 'production' and site.google_analytics.id != '' %}
     <link rel="preconnect" href="https://www.google-analytics.com" crossorigin="use-credentials">
     <link rel="dns-prefetch" href="https://www.google-analytics.com">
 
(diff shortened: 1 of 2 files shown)


cotes2020 (commented):
There is indeed a bug in the loading condition of _Google Analytics_. In addition, other files of the project have some corresponding codes that also need to be modified, but since you are not familiar with Liquid, this PR will be modified after the merge, thank you for your contribution.

PR #541, opened 2022-03-27
Fix various

ZakKemble:
## Description

Some fixes for invalid HTML (using https://validator.w3.org/ to check), pagination fix when using a different path config and fix hidden posts still showing in archives.

## Type of change

- [ ] Documentation update

## How has this been tested

### Test Configuration

- Browser type & version: Firefox 98.0.2
- Operating system: Windows 10
- Ruby version: 2.7.4p191
- Bundler version: 2.3.8
- Jekyll version:  4.2.1

### Checklist

### DIFF
--- a/_includes/post-nav.html
+++ b/_includes/post-nav.html
@@ -9,10 +9,10 @@
     <p>{{ page.previous.title }}</p>
   </a>
   {% else %}
-  <span class="btn btn-outline-primary disabled"
+  <div class="btn btn-outline-primary disabled"
     prompt="{{ site.data.locales[lang].post.button.previous }}">
     <p>-</p>
-  </span>
+  </div>
   {% endif %}
 
   {% if page.next.url %}
@@ -21,10 +21,10 @@
     <p>{{ page.next.title }}</p>
   </a>
   {% else %}
-  <span class="btn btn-outline-primary disabled"
+  <div class="btn btn-outline-primary disabled"
     prompt="{{ site.data.locales[lang].post.button.next }}">
     <p>-</p>
-  </span>
+  </div>
   {% endif %}
 
 </div>
--- a/_includes/post-paginator.html
+++ b/_includes/post-paginator.html
@@ -47,7 +47,7 @@
     {% if show %}
       <!-- show number -->
       <li class="page-item {% if i == paginator.page %} active{% endif %}">
-        <a class="page-link btn-box-shadow" href="{{ site.baseurl }}/{% if i > 1%}page{{ i }}/{% endif %}">{{ i }}</a>
+        <a class="page-link btn-box-shadow" href="{% if i > 1 %}{{ site.paginate_path | replace: ':num', i | relative_url }}{% else %}{{ '/' | relative_url }}{% endif %}">{{ i }}</a>
       </li>
     {% else %}
       <!-- hide number -->
--- a/_includes/refactor-content.html
+++ b/_includes/refactor-content.html
@@ -217,7 +217,7 @@
       {% assign left = left | replace: '">', '"><span class="mr-2">' | append: '</span>' %}
 
       {% assign _new_content = _new_content | append: mark_start
-          | append: left | append: anchor | append: mark_end | append: right
+          | append: left | append: anchor | append: right
       %}
 
     {% endfor %}
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -5,7 +5,7 @@
 <div id="sidebar" class="d-flex flex-column align-items-end">
   <div class="profile-wrapper text-center">
     <div id="avatar">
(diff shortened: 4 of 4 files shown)


_includes/post-nav.html
```diff
@@ -21,10 +21,10 @@
     <p>{{ page.next.title }}</p>
   </a>
   {% else %}
-  <span class="btn btn-outline-primary disabled"
+  <a class="btn btn-outline-primary disabled"
     prompt="{{ site.data.locales[lang].post.button.next }}">
     <p>-</p>
-  </span>
+  </a>
```
cotes2020, line 27: No need for `<a>` here, please use `<div>` instead

_includes/post-paginator.html
```diff
@@ -47,7 +47,7 @@
     {% if show %}
       <!-- show number -->
       <li class="page-item {% if i == paginator.page %} active{% endif %}">
-        <a class="page-link btn-box-shadow" href="{{ site.baseurl }}/{% if i > 1%}page{{ i }}/{% endif %}">{{ i }}</a>
+        <a class="page-link btn-box-shadow" href="{% if i > 1 %}{{ site.paginate_path | replace: ':num', i | relative_url }}{% else %}{{ '/' | relative_url }}{% endif %}">{{ i }}</a>
```
cotes2020, line 50: I don't think `/page{num}/` is an illegal path, please undo the changes here

_layouts/archives.html
```diff
@@ -5,7 +5,8 @@
 
 <div id="archives" class="pl-xl-2">
 
-{% for post in site.posts %}
+{% assign posts = site.posts | where_exp: "item", "item.hidden != true" %}
```
cotes2020, line 8: `hidden: true` only hides the post on the homepage pagination, if we want to hide a certain post site-wide, we should use `published: false` instead. So please revert the changes here.

cotes2020 (changes requested):

_includes/post-paginator.html
ZakKemble: /page{num}/ does not work if "paginate_path" in _config.yml is changed to "/page/:num/".

cotes2020: Is there any particular reason to set it to `/page:num/` in _\_config.yml_? The default one `/page:num/` is good enough.

ZakKemble: I used to use Wordpress which had the pages as `/page/x/` so to keep links and things working I also want to use the same paths in Jekyll.

_layouts/archives.html
ZakKemble: Ah ok

_includes/post-paginator.html
ZakKemble: And of course if someone else wants to change the page paths for whatever reason then it will just work...

_layouts/archives.html
ZakKemble: Actually I'm not sure about this. Using `published` will completly skip generating the page, but what if I want a post that is hidden from the public but can still be accessed if if the URL is known? For this I would use `hidden`, but it still shows up in archives, categories and tags pages which makes it a bit pointless.

_includes/post-paginator.html
cotes2020: Alright, got it

_layouts/archives.html
cotes2020: > what if I want a post that is hidden from the public but can still be accessed if the URL is known?

We could indeed do this by modifying the source code to hide all links to hidden posts on the site, but don't forget that search engines still crawl these hidden posts and display them on the internet. So, what's the point of doing this (`hidden: true`)?

I think the best way to hide a post is to not generate it or encrypt the post content, what do you think?

ZakKemble: If there are no links to the hidden post then search engines will not find it. It will also need to be hidden from feeds and sitemaps. I suppose it does seem like a pain to do properly since other feed and sitemap plugins don't really support hiding things.

cotes2020:
This PR looks a bit chaotic, please make the following changes to the commits:

1. Drop f18d734d3deab29b2c40b49ba1a95dc43faa2ca7 and 07f3a9fa318d157dfc92d6c87de1eda8450382a3 
2. Merge 87743671ad731f7fff22ada67fccd90783c91c5d and 99a0de59ecccafcd61d57f25af4f6218397e68e2 into a single commit.

Each PR should only focus on one issue, too many commits mixed together can make issue tracking difficult, hope to understand. Anyway, thank you very much for your contribution to the project.

cotes2020:
Sorry, I missed one thing: 3ecb2e915481f7d126419b7e016ce985a73e52b8 and 327f5ad26d6020fb377b4ab5a4424ddabb85d205 should also be combined into one commit

PR #590, opened 2022-06-01
Fixes: Some assets is 404 when baseurl isn't empty and assets is 'self_host'

LawssssCat:
## Description

Fixes: Some assets is 404 when baseurl isn't empty and assets is 'self_host'

```yml
baseurl: /i-am-not-empty
assets:
  self_host: true
    enabled:      # boolean, keep empty means false
    # specify the Jekyll environment, empty means both
    # only works if `assets.self_host.enabled` is 'true'
    env:          # [development|production]
```

image

## Type of change

- [ ] Documentation update

## How has this been tested

### Test Configuration

- Browser type & version: chrome
- Operating system: win10
- Ruby version: ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x64-mingw-ucrt] 
- Bundler version: Bundler version 2.3.14 
- Jekyll version:   * jekyll (4.2.2) 

### Checklist

- [ ] My changes generate no new warnings


LawssssCat:
> Thanks for reporting this bug. However, in your patch, cross-domain URLs will also be filtered by `relative_url`, which is not necessary and will affect the readability of the code. So maybe a better way is to handle `baseurl` in `_data/assets/self_host.yml`

It may affect readability, but this `relative_url` is officially provided by liquid, so everyone should be aware of its function, and the liquid seems to have a hunch that `relative_url` will be uesed to filter `cross-domain URLs` so it doesn't seem to do extra processing for `cross-domain URLs`. On the contrary, every time you modify the `baseurl`, you need to modify `_data/assets/self_host.yml` synchronously, which makes me fell very troublesome, so I think `relative_url` should be added.

LawssssCat:
> Maybe you don't get my point, of course we need `relative_url`, but I mean it should be used in the file `_data/assets/self_host.yml`, not in those HTML files.

Yes， I didn't understand what you were trying to du, because I just tried it and `relative_url` doesn't seem to work in `_data/*`
(Is there any syntax I don't know yet?)

```yml
---
---

# fonts

webfonts: {{ "/assets/lib/fonts/main.css" | relative_url }}
```

cotes2020:
Well, I didn't notice that Jekyll doesn't support parsing Liquid filters in data files, so adding baseurl in the HTML file is the only option left, will merge this PR.

_includes/js-selector.html
```diff
@@ -36,10 +36,10 @@
...
 
   {% assign _urls = site.data.assets[origin].dayjs.js.common
-      | append: ',' | append: site.data.assets[origin].dayjs.js.locale
-      | replace: ':LOCALE', locale
-      | append: ',' | append: site.data.assets[origin].dayjs.js.relativeTime
-      | append: ',' | append: site.data.assets[origin].dayjs.js.localizedFormat
+    | append: ',' | append: site.data.assets[origin].dayjs.js.locale
+    | replace: ':LOCALE', locale
+    | append: ',' | append: site.data.assets[origin].dayjs.js.relativeTime
+    | append: ',' | append: site.data.assets[origin].dayjs.js.localizedFormat
```
cotes2020, line 42: Please undo these indents

cotes2020 (changes requested):

cotes2020 (changes requested):

cotes2020 (changes requested):
If you can, squash all your commits, then rebase to the latest commit on the master branch.

PR #598, opened 2022-06-10
fix: typo in Hungarian locale file

borosbence:
## Description

In the 'mighty' hungarian language we don't use plurals after numbers. This will fix it on the Categories page.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## How has this been tested

### Test Configuration

- Browser type & version:
- Operating system:
- Ruby version: 
- Bundler version: 
- Jekyll version: 

### Checklist

### DIFF
--- a/_data/locales/hu-HU.yml
+++ b/_data/locales/hu-HU.yml
@@ -77,9 +77,5 @@ post:
 
 # categories page
 categories:
-  category_measure:
-    singular: kategória
-    plural: kategória
-  post_measure:
-    singular: bejegyzés
-    plural: bejegyzés
+  category_measure: kategória
+  post_measure: bejegyzés
--- a/_data/locales/hu-HU.yml
+++ b/_data/locales/hu-HU.yml
@@ -79,7 +79,7 @@ post:
 categories:
   category_measure:
     singular: kategória
-    plural: kategóriák
+    plural: kategória
   post_measure:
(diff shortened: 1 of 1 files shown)


cotes2020 (commented):
Are you sure that the singular and plural of these two words in Hungarian are the same? Because you just submitted #597 an hour ago

borosbence:
Their plural form is different from singular, but if they appear after a number, they are grammatically correct. Unfortunately, I didn’t check out these two new texts in the categories at first.

_data/locales/hu-HU.yml
```diff
@@ -79,7 +79,7 @@ post:
 categories:
   category_measure:
     singular: kategória
-    plural: kategóriák
+    plural: kategória
   post_measure:
     singular: bejegyzés
-    plural: bejegyzések
+    plural: bejegyzés
```
cotes2020, line 85: If there is no need to distinguish between singular and plural forms, it can be modified as follows:

```yaml
categories:
  category_measure: kategória
  post_measure: bejegyzés
```

cotes2020 (changes requested):

_data/locales/hu-HU.yml
borosbence: Thanks! That made it better.

PR #833, opened 2023-01-01
feat(i18n) add Czech language

MatyIRG:
I made Czech translation.

### DIFF
--- a/_data/locales/cs-CZ.yml
+++ b/_data/locales/cs-CZ.yml
@@ -1,0 +1,93 @@
+# The layout text of site
+
+# ----- Commons label -----
+
+layout:
+  post: Příspěvek
+  category: Kategorie
+  tag: Štítek
+
+# The tabs of sidebar
+tabs:
+  # format: <filename_without_extension>: <value>
+  home: Domů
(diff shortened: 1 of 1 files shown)


_data/locales/cs-CZ.yml
```diff
@@ -0,0 +1,93 @@
...
+    dayjs: 'll'
+  archives:
+    strftime: '%b'
+    dayjs: 'MMM'
+
+# categories page
+categories:
+  category_measure:
+    singular: kategorie
+    plural: kategorie
```
cotes2020, line 90: If singular and plural are the same, you can use `category_measure ` to define it:

```yml
# ...
categories:
  category_measure: kategorie
# ...
```

cotes2020 (changes requested):

_data/locales/cs-CZ.yml
MatyIRG: Yes, they are same. I will change it.

MatyIRG (commented):
Changed

cotes2020 (approved):

PR #965, opened 2023-04-05
Fix a bug in feed.xml

v4vong:
## Description

It might cause a syntax error if the `post.title` contains double quotes. One of the solutions is using the `escape` filter.

## Type of change

- [ ] Documentation update

## How has this been tested

### Test Configuration

- Browser type & version:
- Operating system:
- Ruby version: 
- Bundler version: 
- Jekyll version:

### DIFF
--- a/assets/feed.xml
+++ b/assets/feed.xml
@@ -26,7 +26,7 @@ permalink: /feed.xml
   {% assign post_absolute_url = post.url | absolute_url %}
   <entry>
     <title>{{ post.title }}</title>
-    <link href="{{ post_absolute_url }}" rel="alternate" type="text/html" title="{{ post.title }}" />
+    <link href="{{ post_absolute_url }}" rel="alternate" type="text/html" title="{{ post.title | escape }}" />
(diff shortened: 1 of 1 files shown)


assets/feed.xml
```diff
@@ -26,7 +26,7 @@ permalink: /feed.xml
   {% assign post_absolute_url = post.url | absolute_url %}
   <entry>
     <title>{{ post.title }}</title>
-    <link href="{{ post_absolute_url }}" rel="alternate" type="text/html" title="{{ post.title }}" />
+    <link href="{{ post_absolute_url }}" rel="alternate" type="text/html" title="{{ post.title | escape }}" />
```
cotes2020, line 29: Thank you for fixing it. Using the filter provided by Jekyll might be better.

```suggestion
    <link href="{{ post_absolute_url }}" rel="alternate" type="text/html" title="{{ post.title | xml_escape }}" />
```

cotes2020 (changes requested):

assets/feed.xml
v4vong: Great

PR #1015, opened 2023-05-01
feat: allow users to define custom metadata

v4vong:
## Description

This new feature allows users to define custom metadata in the `<head>` without overriding the whole `head.html`. They just need to put their metadata into `custom-head.html`, then the metadata would be automatically included into the `<head>`. 

## Type of change

- [ ] Documentation update

## How has this been tested

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -113,4 +113,7 @@
   {% unless site.theme_mode %}
     {% include mode-toggle.html %}
   {% endunless %}
+
+  {% include metadata-hook.html %}
+
 </head>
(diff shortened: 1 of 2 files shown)


cotes2020 (changes requested):
It would be better to rename `custom-head.html` to `metadata-hook.html`.

_includes/custom-head.html
```diff
@@ -0,0 +1,3 @@
+{% comment %}
+  A placeholder to allow defining custom metadata
+{% endcomment %}
```
cotes2020, line 3:
```suggestion
<!-- A placeholder to allow defining custom metadata -->
```
Comments using HTML style will be more compact.

cotes2020 (changes requested):

_includes/custom-head.html
v4vong: Accepted

PR #1105, opened 2023-06-25
The CJK characters of the "Search Cancel" button will wrap

avennn:
## Description

H5 search cancel text arranges vertically when lang switched to zh-CN. I think it is a bug.

## Type of change

## How has this been tested

- follow the [contribution doc](https://github.com/cotes2020/jekyll-theme-chirpy/blob/master/.github/CONTRIBUTING.md)，fork the project and clone it locally
- install deps
- change`_config.yml`: `lang: en` to `lang: zh-CN`
- run `npm run build && bundle exec jekyll serve`
- view in chrome devtool with mobile mode, click search icon, you will see "取消" arranging vertically at right side
- with this commit, reproduce above steps and get the correct render result.

- [x] I have run `bash ./tools/test` (at the root of the project) locally and passed

### DIFF
--- a/_sass/addon/commons.scss
+++ b/_sass/addon/commons.scss
@@ -957,7 +957,7 @@ $btn-mb: 0.5rem;
 /* 'Cancel' link */
 #search-cancel {
   color: var(--link-color);
-  margin-left: 1rem;
+  margin-left: 0.75rem;
   display: none;
   white-space: nowrap;
 
--- a/_sass/addon/commons.scss
+++ b/_sass/addon/commons.scss
@@ -959,6 +959,7 @@ $btn-mb: 0.5rem;
   color: var(--link-color);
   margin-left: 1rem;
   display: none;
+  white-space: nowrap;
 
   @extend %cursor-pointer;
 }


cotes2020 (changes requested):
Thanks for raising this issue! But I think there is a better solution.

_sass/addon/commons.scss
```diff
@@ -959,6 +959,7 @@ $btn-mb: 0.5rem;
   color: var(--link-color);
   margin-left: 1rem;
   display: none;
+  white-space: nowrap;
```
cotes2020, line 962: 1. Reserve `4rem` space for the cancel button and revert the `white-space` here
```suggestion
  width: 4rem;
  display: none;
```

2. Add BS class for the button to handle the wrap issue:
https://github.com/cotes2020/jekyll-theme-chirpy/blob/74ab6f8adc2a7930e7bf21543ad17ded386c93c3/_includes/topbar.html#L68

```diff
-    <span id="search-cancel">{{ site.data.locales[include.lang].search.cancel }}</span>
+    <span id="search-cancel" class="text-truncate text-center">
+      {{- site.data.locales[include.lang].search.cancel -}}
+    </span>
```

cotes2020 (changes requested):

_sass/addon/commons.scss
avennn: oh, you want overflow text ellipsis. But i think this code still has problem:

1. cancel text too short, too much space at the right side
image
2. cancel text too long, no space between cancel text and input
image

_sass/addon/commons.scss
```diff
@@ -959,6 +959,7 @@ $btn-mb: 0.5rem;
   color: var(--link-color);
   margin-left: 1rem;
```
cotes2020, line 960: Alright, please ignore the modification request above. Keep your changes, and add a margin improvement:

```suggestion
  margin-left: 0.75rem;
```

cotes2020 (changes requested):

_sass/addon/commons.scss
avennn: ok

cotes2020 (approved):

PR #1130, opened 2023-07-06
Update commons.scss icon for prompt-info

khoguan:
## Description

Both the icons for .prompt-info and .promt-warning are '\f06a'. But the icon for .prompt-info should be '\f05a' (circle-info from fontawesome), not '\f06a' (circle-exclamation).

## Type of change


cotes2020 (commented):
The icon was chosen based on how it looks, not the name.

PR #1265, opened 2023-09-24
German language localisation

jensboe:
* Add common German date format.
* Fix translation of read time prompt to a more correct word.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] Improvement (refactoring and improving code)

## Description
* Add common German date formats.
* Fix translation of read time prompt to a more correct word.

## Additional context
Read time unit needs a singular and plural version to be correctly translated.

### DIFF
--- a/_data/locales/de-DE.yml
+++ b/_data/locales/de-DE.yml
@@ -75,11 +75,8 @@ post:
 df:
   post:
     strftime: "%d.%m.%Y"
-    dayjs: "D.M.YYYY"
-  archives:
-    strftime: "%l"
-    dayjs: "M"
-
+    dayjs: "DD.MM.YYYY"
+    
 # categories page
 categories:
   category_measure:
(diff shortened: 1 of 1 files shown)


_data/locales/de-DE.yml
```diff
@@ -70,6 +70,16 @@ post:
...
+# Date time format.
+# See: <http://strftime.net/>, <https://day.js.org/docs/en/display/format>
+df:
+  post:
+    strftime: "%d.%m.%Y"
+    dayjs: "D.M.YYYY"
+  archives:
+    strftime: "%l"
+    dayjs: "M"
+
```
cotes2020, line 82:
```suggestion
```

Please resume these

cotes2020 (changes requested):

_data/locales/de-DE.yml
cotes2020: The output of `dayjs` must be the same as `strftime`, otherwise it will jitter the text when loading.

The month of `df.archives` is wrong, you set it to the hour (`%l`), when submitting PR, please test it locally, CI can't check this kind of text content change error.

PR #1440, opened 2023-12-30
Fix htmlproofer path in ci

karlmolina:
## Type of change

- [ ] Documentation update

## Description

I used this repo as a template and the github actions failed because I guess htmlproofer was looking at the wrong directory. Once I changed `_site` to `_site/blog` it fixed the issue.
Here is the failing action: https://github.com/karlmolina/blog/actions/runs/7366373473/job/20048587742#step:6:475
And here is the fixed action after my change: https://github.com/karlmolina/blog/actions/runs/7366433467/job/20048693278

If this is supposed to work this way for some reason let me know and you can close the pr.

### DIFF
--- a/.github/workflows/pages-deploy.yml.hook
+++ b/.github/workflows/pages-deploy.yml.hook
@@ -52,7 +52,7 @@ jobs:
 
       - name: Test site
         run: |
-          bundle exec htmlproofer _site \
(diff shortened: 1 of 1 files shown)


karlmolina:
Nevermind I realized I need to set the baseurl in _config.yml to `/blog` so all the links have that prefix.

Issue #1532, opened 2024-02-08
PageView counter at top of the posts using Goatcounter


sukhjinder-kumar:
### Checklist

- [X] I have read and accepted the [contributing guidelines](https://github.com/cotes2020/jekyll-theme-chirpy/blob/master/docs/CONTRIBUTING.md).

### Is your feature request related to a problem? Please describe

Add page view count on top of the blog post.

### Describe the solution you'd like

Recently support for goatcounter has been provided. As per this [documentation](https://www.goatcounter.com/help/visitor-counter#:~:text=You%20can%20display%20a%20page's,prevent%20unintentional%20leaking%20of%20data.&text=You%20can%20also%20remove%20the%20async%20attribute.) it is very much possible to print page view on our posts.

We can call the API and get the page count as follows -
```js
// Fetch JSON data from the provided URL
fetch('https://[MYCode].goatcounter.com/counter/[relative URL].json')
  .then(response => response.json())
  .then(data => {
    console.log(data.count);
  })
  .catch(error => console.error('Error fetching JSON:', error));
```
I am not very good at web development, and hence I am not able to integrate it with the templates. Even adding to the index.html page gives errors (`Error fetching Json`). 
```html
<div id="jsonAttribute"></div>

<script>
  // Fetch JSON data from the provided URL
  fetch('https://[MYCode].goatcounter.com/counter/[relative URL].json')
    .then(response => {response.json(); console.log(response.json());})
    .then(data => {
      // Display the desired attribute
      document.getElementById('jsonAttribute').innerText = data.count;
    })
    .catch(error => console.error('Error fetching JSON:', error));
</script>
```

If it helps, Chatgpt 3.5 states the following as a possible error "If you're able to fetch the JSON data successfully using the same JavaScript code in a .js file when running it with Node.js but encountering an error when trying to fetch it within an HTML file, the issue might be related to CORS (Cross-Origin Resource Sharing) restrictions."

From my understanding of the chirpy template, one has to add a working HTML code (shown above) in `post.html`. And add some CSS to make it look clean. Maybe display it to the right, alongside posted and updated (at the start of the post), as in below.
![Screenshot from 2024-02-09 04-49-09](https://github.com/cotes2020/jekyll-theme-chirpy/assets/82507694/698c10de-9f0c-4098-860b-b3efbd29ccc2)

kungfux:
[...]

PR #1543, opened 2024-02-16
feat: add post views counter for GoatCounter

kungfux:
## Type of change

## Description
I've integrated and use GoatCounter for a while in my own codebase and since GoatCounter have been integrated into Chirpy, I'd like to incorporate changes for the post views counter. Other people also want this feature according to ref #1532.

## Implementation
First, we need to connect GoatCounter in the config, and then enable views counter by specifying the source provider for it like it is done for comments. This is how it looks like in the UI currently.

image

## Ideas
- Since it takes time to retrieve the data from counter provider, I'm not happy to show "0 views" initially and so I may suggest to hide views counter initially and then display it once actual views number is available by applying `visibility: hidden` and `visibility: visible` styles to a parent `span` element. Here is an [example](https://github.com/kungfux/kungfux.github.io/commit/ac2aee9603e32a671f8dd123e095381970eeba7a).

## Additional context
Since I was surprised to find localized string `pageview_measure`, I suspect you may have other plans for the same feature in mind, or have passed similar way in the past that you can share. Anyways, let me know what do you think once you have a time. Thanks.


cotes2020 (changes requested):
Thank you so much! While I didn't actually use a GoatCounter ID to test the code, I'm sure it's available. Also, I have a couple ideas to share with you:

1. For the UI design, I think the location of the view count would be more appropriate to the right of the reading time.

    pv


2. The reference to the JS of GoatCounter pageviews should be placed in `_includes/js-selector.html`, which will bring better web page rendering performance.

3. The number to initialize the view count could be set to `1` instead of `0`.

4. i18n's `pageview_measure` was once used for Google Analaytics pageviews reporting, this feature was removed in #1071 as UA was deprecated. i18n keeps it for future pageviews services (like goatcounter now).

cotes2020:
> 1. For the UI design, I think the location of the view count would be more appropriate to the right of the reading time.
> 
>    pv

I'd like to change the above comment: placing views to the left of the reading time is the way to go, otherwise it will cause jitters in the nearby text (reading time) due to JS modifying the number.

Updated:

pv-edited

kungfux:
Thank you for taking the time to review the suggested changes and for sharing great ideas.

I've made suggested changes for UI and JS injection location. Please take a quick look. Thank you very much.

cotes2020 (commented):
Since we now have multiple web analytics options, wouldn't it be a good idea to improve the configured options to make it more compact similar to the comment system.

`_config.yml`:

```yaml
# Web Analytics Settings
analytics:
  provider: # [google | goatcounter]
  google:
    id: # fill in your Google Analytics ID
  goatcounter:
    id: # fill in your GoatCounter ID
    pv:
      enabled: # boolean, the option for the post views counter, keep empty means disable
```

`_includes/` folder:

```
_includes/
└── analytics
    ├── goatcounter.html
    └── google.html
```

But doing so would require refactoring the existing google analytics code, what do you think?

kungfux:
Sure, I like the idea of merging configuration settings to have everything in a single configuration section.

The only thing I would suggest changing is to allow using both analytics platforms at the same time.

However, I wonder how breaking configuration changes will be released? Will it be part of a major release or will changes need to be back-compatible? How do you usually do?

cotes2020:
I favor the idea of using multiple analytics platforms at the same time, in which case the settings would have to be designed as:

```yaml
## _config.yml

# Web Analytics Settings
analytics:
  google:
    id: # fill in your Google Analytics ID
  goatcounter:
    id: # fill in your GoatCounter ID
```

Since pageviews can only display data from a single platform, the config options would be:

```yaml
## _config.yml

# Pageviews settings
pageviews:
  provider: # [google | goatcounter]
```

Generally speaking, breaking changes should be released in the MAJOR version, and strictly enforcing this guideline would result in too many major releases. So for this feature, I prefer to release it on the MINOR version, and remind users of the changes in the release note.

kungfux:
Sounds good. Thanks.

Configuration changes have been implemented and seems to work as expected. Please have a look.

cotes2020:
I've done a lot of refactoring, so if there are any thoughts on this, please let me know.

kungfux:
The only thought is: It's nicely done 👍

I have enjoyed the solution with the spinner 🤩 and learnt something new. Thank you.

cotes2020 (approved):

PR #1559, opened 2024-02-25
build(deps-dev): remove deprecated library and add new one the npm group with 1 update

seod0209:
## Description
- @babel/plugin-proposal-class-properties was deprecated
  - https://www.npmjs.com/package/@babel/plugin-proposal-class-properties
- update to use @babel/plugin-transform-class-properties instead


cotes2020 (approved):
Thanks 👍

PR #1586, opened 2024-03-02
feat(i18n): allow `page.lang` to override `site.lang`

neilboyd:
## Type of change

## Description
Allow `lang` on page/post to override `site.lang`.

This is useful to create folders (collections) of different languages. For example, I have folders `_en`, `_ru` and `_tr` (relative urls `en`, `ru` and `tr`), and then define these as collections in `_config.xml`:
```yaml
collections:
  en:
    output: true
  ru:
    output: true
  tr:
    output: true

defaults:
  - scope:
      path: ""
      type: en
    values:
      lang: en
  - scope:
      path: ""
      type: tr
    values:
      lang: tr-TR
  - scope:
      path: ""
      type: ru
    values:
      lang: ru-RU
```

It's also possible to just put `lang: ru-RU` directly into the page front matter.

## Additional context
I removed usage of `site.alt_lang` in the `html lang` tag of `default` layout because I'm not sure what that was for. If there's a reason for it then I could put it back.

I don't exactly know what `js-selector.html` is doing with `lang`, so I couldn't test it, but the change seems obvious.

### DIFF
--- a/_includes/js-selector.html
+++ b/_includes/js-selector.html
@@ -33,7 +33,7 @@
   or page.layout == 'category'
   or page.layout == 'tag'
 %}
-  {% assign locale = site.lang | split: '-' | first %}
+  {% assign locale = include.lang | split: '-' | first %}
 
   {% assign urls = urls
     | append: ','
--- a/_includes/js-selector.html
+++ b/_includes/js-selector.html
@@ -33,7 +33,7 @@
   or page.layout == 'category'
   or page.layout == 'tag'
 %}
-  {% assign locale = page.lang | default: site.lang | split: '-' | first %}
+  {% assign locale = include.lang | split: '-' | first %}
 
   {% assign urls = urls
     | append: ','
--- a/_includes/js-selector.html
+++ b/_includes/js-selector.html
@@ -33,7 +33,7 @@
   or page.layout == 'category'
   or page.layout == 'tag'
 %}
-  {% assign locale = include.lang | split: '-' | first %}
+  {% assign locale = page.lang | default: site.lang | split: '-' | first %}
 
   {% assign urls = urls
     | append: ','
--- a/_includes/lang.html
+++ b/_includes/lang.html
@@ -1,10 +1,8 @@
 {% comment %}
   Detect appearance language and return it through variable "lang"
 {% endcomment %}
-{% if site.data.locales[page.lang] %}
-  {% assign lang = page.lang %}
-{% else if site.data.locales[site.lang] %}
-  {% assign lang = site.lang %}
-{% else %}
+{% assign lang = page.lang | default: site.lang %}
+
+{% unless site.data.locales[lang] %}
   {% assign lang = 'en' %}
-{% endif %}
+{% endunless %}
--- a/_includes/lang.html
+++ b/_includes/lang.html
@@ -1,8 +1,10 @@
 {% comment %}
   Detect appearance language and return it through variable "lang"
 {% endcomment %}
-{% assign lang = page.lang | default: site.lang %}
-
-{% unless site.data.locales[lang] %}
+{% if site.data.locales[page.lang] %}
+  {% assign lang = page.lang %}
+{% elsif site.data.locales[site.lang] %}
+  {% assign lang = site.lang %}
+{% else %}
   {% assign lang = 'en' %}
-{% endunless %}
+{% endif %}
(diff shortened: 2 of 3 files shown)


cotes2020 (commented):
Thanks for adding this feature. I understand your motivation and it makes sense.

- `site.alt_lang` is a trick to get the site to use a different `lang` attribute than the language on the UI.

- The `lang` of the `js-selector.html` is used to specify the datetime format of the tooltip, which is what is inside the floating window when the mouse floats to the date.
   Screen Shot 2024-03-03 at 15 47 47

_includes/lang.html
```diff
@@ -1,7 +1,9 @@
 {% comment %}
   Detect appearance language and return it through variable "lang"
 {% endcomment %}
-{% if site.data.locales[site.lang] %}
+{% if site.data.locales[page.lang] %}
+  {% assign lang = page.lang %}
+{% else if site.data.locales[site.lang] %}
   {% assign lang = site.lang %}
 {% else %}
   {% assign lang = 'en' %}
```
cotes2020, line 9:
```diff
-{% if site.data.locales[site.lang] %}
-  {% assign lang = site.lang %}
-{% else %}
+{% assign lang = page.lang | default: site.lang %}
+
+{% unless site.data.locales[lang] %}
   {% assign lang = 'en' %}
-{% endif %}
+{% endunless %}
```

cotes2020: The `lang.html` can be optimized.

_includes/js-selector.html
```diff
@@ -33,7 +33,7 @@
   or page.layout == 'category'
   or page.layout == 'tag'
 %}
-  {% assign locale = site.lang | split: '-' | first %}
+  {% assign locale = include.lang | split: '-' | first %}
```
cotes2020, line 36:
```suggestion
  {% assign locale = page.lang | default: site.lang | split: '-' | first %}
```

cotes2020 (changes requested):

_layouts/default.html
```diff
@@ -12,8 +12,7 @@
   {% capture prefer_mode %}data-mode="{{ site.theme_mode }}"{% endcapture %}
 {% endif %}
 
-<!-- `site.alt_lang` can specify a language different from the UI -->
-<html lang="{{ site.alt_lang | default: site.lang }}" {{ prefer_mode }}>
+<html lang="{{ lang }}" {{ prefer_mode }}>
```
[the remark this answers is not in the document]
neilboyd: But then it's not using `page.lang`. How about?
```
<html lang="{{ site.alt_lang | default: page.lang | default: site.lang }}" {{ prefer_mode }}>
```

I still don't understand why you'd want an alt lang.

_includes/lang.html
neilboyd: This logic is slightly different - if the page lang is invalid then my way would use `site.lang`, whereas this way will use `en`. But either argument is valid, so simpler is better!

_layouts/default.html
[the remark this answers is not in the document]
cotes2020: Let's put it this way, the `lang` attribute in `<html>` is used to indicate the language of the global content, but if I want the basic UI elements (like tabs) to be in English and the post content to be in French, then I can set `site.lang=en` as well as `site.alt_lang=fr`.

- Specify UI language: `page.lang`, `site.lang`
- Specify write content language: `site.alt_lang`

cotes2020, line 15: In a way `page.lang` and `site.alt_lang` serve a similar purpose, but `page.lang` should have a higher priority than `site.alt_lang`, so I now think it should be changed to something like the following:

```suggestion
<!-- `site.alt_lang` can specify a language different from the UI -->
	<html lang="{{ page.lang | default: site.alt_lang | default: site.lang }}" {{ prefer_mode }}>
```

_includes/lang.html
cotes2020: > This logic is slightly different - ...

You're right, so please ignore my advice and maintain the changes you submitted the first time.

```diff
-{% if site.data.locales[site.lang] %}
+{% if site.data.locales[page.lang] %}
+  {% assign lang = page.lang %}
+{% else if site.data.locales[site.lang] %}
   {% assign lang = site.lang %}
 {% else %}
   {% assign lang = 'en' %}
```

cotes2020 (changes requested):

cotes2020 (approved):

Issue #1596, opened 2024-03-07
Liquid code being displayed in related-posts.html


steveng57:
### Checklist

- [X] I have read and accepted the [contributing guidelines](https://github.com/cotes2020/jekyll-theme-chirpy/blob/master/docs/CONTRIBUTING.md).

### How did you create the site?

Generated from `chirpy-starter`

### Describe the bug

In my post markdown, if I have any liquid code immediately after the Front Matter, when this post is displayed in the "Further Reading" section (which I believe is defined by related-posts.html), then the liquid code is being displayed, instead of it being stripped out.

### Steps To Reproduce

In related-posts.html the code to display a post is below, from line 85 or so.  There doesn't look to be anything to strip out Liquid code.
```liquid
                  {% include no-linenos.html content=post.content %}
                  {{ content | markdownify | strip_html | truncate: 200 | escape }}
```

### Expected Behavior

I'm not a Jekyll expert by any means, but it looks like the code should be...
```liquid
                  {% capture new-content %}
                    {% include no-linenos.html content=post.content %}
                  {% endcapture %}
                  {{ new-content | markdownify | strip_html | truncate: 200 | escape }}
```

### Environment

- Ruby:  ruby 3.2.1 (2023-02-08 revision 31819e82c8) [x64-mingw-ucrt]
- Jekyll:  jekyll 4.3.3
- Chirpy:  jekyll-theme-chirpy (6.5.3)

steveng57:
Another, perhaps better solution, is to look for a 'description' field in the post Front Matter.  If it does not exist, carry on as usual, if it does exist however, use it in the related-posts (Further Reading) sections, and also in the descriptive blurb in the home page.

kungfux:
I think it's a great feature to have a `description` field in the front matter, as it can help work around this problem as well as https://github.com/cotes2020/jekyll-theme-chirpy/discussions/1516. However, the original issue should be fixed separately.

I would like to have this feature for my posts as well, to have control over the post description that appears in multiple places. I will work on it.

cotes2020:
Which version of Chirpy are you using? Please fill in the software environment or this issue will be considered an invalid report, thank you.

kungfux:
I'm able to reproduce the problem reported on latest `master` branch. While @steveng57 is using custom HTML file to include some `<head>`'s, I have used video embedding to demo it below.

```markdown
---
title: Writing a New Post
author: cotes
[...]

PR #1602, opened 2024-03-10
feat: make post description customizable

kungfux:
## Type of change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Description
I would like to have control over the post summary that appears on the `Home` page, in the `Further Reading` section, and in the RSS feed instead of the auto-generated one that is not always great for this purpose.

ref #1596
ref #1516 

## BREAKING CHANGE
A similar feature `summary` already exists and is used for RSS feed summary exclusively. However, it's support is removed with this PR and replaced by `description`.

## Renderings

- Homepage
  Screenshot 2024-03-18 at 18 40 43

- Posts page
  Screenshot 2024-03-18 at 18 41 11

### DIFF
--- a/_data/origin/cors.yml
+++ b/_data/origin/cors.yml
@@ -13,7 +13,7 @@ cdns:
 
 # fonts
 
-webfonts: https://fonts.googleapis.com/css2?family=Lato&family=Source+Sans+Pro:wght@400;600;700;900&display=swap
+webfonts: https://fonts.googleapis.com/css2?family=Lato:wght@300;400&family=Source+Sans+Pro:wght@400;600;700;900&display=swap
 
 # Libraries
 
--- a/_includes/post-description.html
+++ b/_includes/post-description.html
@@ -1,12 +1,16 @@
-{% comment %}
+{%- comment -%}
   Get post description or generate it from the post content.
-{% endcomment %}
+{%- endcomment -%}
 
-{% assign max_length = include.max_length | default: 200 %}
+{%- assign max_length = include.max_length | default: 200 -%}
 
-{% if post.description %}
-  {{ post.description | strip | truncate: max_length }}
-{% else %}
-  {% include no-linenos.html content=post.content %}
-  {{ content | markdownify | strip_html | truncate: max_length | escape }}
-{% endif %}
+{%- capture description -%}
+{%- if post.description -%}
+  {{- post.description -}}
+{%- else -%}
+  {%- include no-linenos.html content=post.content -%}
+  {{- content | markdownify | strip_html -}}
+{%- endif -%}
+{%- endcapture -%}
+
+{{- description | strip | truncate: max_length | escape -}}
--- a/_includes/post-description.html
+++ b/_includes/post-description.html
@@ -3,8 +3,9 @@
 {% endcomment %}
 
 {% assign max_length = include.max_length | default: 200 %}
+
 {% if post.description %}
-  {{ post.description | strip | truncate: max_length | escape }}
+  {{ post.description | strip | truncate: max_length }}
 {% else %}
   {% include no-linenos.html content=post.content %}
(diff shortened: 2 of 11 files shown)


cotes2020 (commented):
Looks like a nice feature, thank you @kungfux! 

I think using a `description` field would be a better choice because the plugin `jekyll-seo-tag` recognizes the `page.description` instead of `page.summary`.

kungfux:
> I think using a `description` field would be a better choice because the plugin `jekyll-seo-tag` recognizes the `page.description` instead of `page.summary`.

I see. It's a good point. Thanks @cotes2020 

I have updated PR accordingly.

To avoid breaking changes, the support of `summary` field remains in place and will be utilized for RSS feed exclusively when defined.

cotes2020:
Until `7.0.0` is released, feel free to submit breaking changes, the `page.summary` is simply defined according to the atom feed tag `<summary>`.

kungfux:
> Until `7.0.0` is released, feel free to submit breaking changes, the `page.summary` is simply defined according to the atom feed tag `<summary>`.

Ok, I have removed support of `summary` field to unify behavior across multiple places and marked PR with breaking changes "just in case".

cotes2020 (commented):
So I've done some refactoring that

- Fixed the formatting
- Changed the word limit (`max_length`)
- Show the  `page.description` under the post title 

If you have any ideas, feel free to leave a comment.

kungfux:
> Fixed the formatting

Not sure why there are differences between our dev environments as Prettier is configured on my end and it makes changes that you revert.

> Changed the word limit (max_length)

I think it might be a good idea to make limit customizable for RSS feed from `_config.yml` (could be a separate PR). If you don't force people to visit your blog, limiting it to `400` characters kills the idea of using reader applications.

> Show the `page.description` under the post title

This makes a lot of sense 👍

cotes2020:
### Formatting issues

I use VSCode, and the basic configuration is in [`.vscode`](https://github.com/cotes2020/jekyll-theme-chirpy/tree/master/.vscode). The extension `Shopify.theme-check-vscode` deletes a blank line immediately after the front-matter when formatting, so to avoid this, my trick is to edit the file with two blank lines at the bottom of the second `---` of the front-matter, so that the auto-formatting will leave only one blank line.

### Word limit for posts in RSS feed

> I think it might be a good idea to make limit customizable for RSS feed from `_config.yml` (could be a separate PR). If you don't force people to visit your blog, limiting it to `400` characters kills the idea of using reader applications.

According to Atom's definition of [`<summary>`](https://validator.w3.org/feed/docs/atom.html#recommendedEntryElements), its content is excerpted text, not the complete content. So I don't think it's appropriate to use a length of `1000`, and there's also no need to customize the length of the `<summary>` in `_config.yml`.

kungfux:
> Formatting issues

I see where the problem is. If we want to force `Shopify.theme-check-vscode` to follow this formatting, then we may need to build custom rule, it seems.

> Word limit for posts in RSS feed

Thanks for insights @cotes2020. You're right. I have to work with the `content` field and it is a matter of RSS reader to support the format and display the content followed by `content.src`. We are good here for now.

cotes2020 (approved):
OK, that's it :)

dineshkoravi:
Hi @kungfux 
are the changes merged ? i can see the code updated in master, but changes are not visible in demo page. This is a neat feature and would like to check this. I have my site with latest version and not able to use this feature. can you help please ?

kungfux:
> Hi @kungfux 
> are the changes merged ? i can see the code updated in master, but changes are not visible in demo page. This is a neat feature and would like to check this. I have my site with latest version and not able to use this feature. can you help please ?

Hey @dineshkoravi 
The feature is merged but not released yet. I wait for it as well 🤞

PR #1611, opened 2024-03-16
feat(footer): display theme version

MrMurdock11:
## Type of change

## Description
Summary: Display theme version.
Motivation: When I'm working on my blog, I like to see the current version of the theme I'm using.
 - It's helpful to see which version we use, it helps to avoid unnecessary steps to find the theme version.
 - It also makes it easier to write reports, as users can see the version they have a problem with right away. Especially if the user specifies theme like this: ">= 5.6".

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -34,7 +34,7 @@
     {%- endcapture -%}
 
     {%- capture _theme -%}
-      <a href="https://github.com/cotes2020/jekyll-theme-chirpy" target="_blank" rel="noopener">Chirpy</a>
+      <a href="https://github.com/cotes2020/jekyll-theme-chirpy" target="_blank" rel="noopener">Chirpy@{{ theme.version }}</a>
     {%- endcapture -%}
 
     {{ site.data.locales[include.lang].meta | replace: ':PLATFORM', _platform | replace: ':THEME', _theme }}
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -34,7 +34,14 @@
     {%- endcapture -%}
 
     {%- capture _theme -%}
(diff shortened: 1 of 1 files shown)


_includes/footer.html
```diff
@@ -34,7 +34,7 @@
     {%- endcapture -%}
 
     {%- capture _theme -%}
-      <a href="https://github.com/cotes2020/jekyll-theme-chirpy" target="_blank" rel="noopener">Chirpy</a>
+      <a href="https://github.com/cotes2020/jekyll-theme-chirpy" target="_blank" rel="noopener">Chirpy@{{ theme.version }}</a>
```
GetOutOfMyBakery, line 37: Drive-by comment, because I really like this idea and I've just been messing about with it locally myself.

How about only displaying it as a tooltip though?
```suggestion
      <span
        data-bs-toggle="tooltip"
        data-bs-placement="top"
        title="Version: {{ theme.version }}"
      >
        <a href="https://github.com/cotes2020/jekyll-theme-chirpy" target="_blank" rel="noopener">Chirpy</a>
      </span>
```

image

cotes2020: @GetOutOfMyBakery I vote for your idea.

MrMurdock11: Yes, I like this option as well. In this case the information will be less explicit, but at least it won't be out of the old display format. The question now is which option is better: Implicit or Explicit.

What do you think about shortening it to "v6.5.3" and not using the `span` wrapper?
image

cotes2020: Implicit and `vX.Y.Z` @MrMurdock11

cotes2020 (approved):
LGTM

MrMurdock11:
@cotes2020 I can't merge this changes

PR #1613, opened 2024-03-16
feat: add 'jump to main content' button

GetOutOfMyBakery:
## Type of change

- [ ] Documentation update

## Description

Adds a configurable `Jump to main content` hidden button that is displayed only when focused by pressing tab.

This makes it easier to navigate the blog just using a keyboard or screen reader.
1. Navigate to your blog
1. Press tab to go to the first tab stop: `Jump to main content`
1. Press enter, and the focus switches to the `<main>` element. Subsequent `tab`s will now be within the body of the `home`/`post`/`categories`/`tags`/`archive`/`about` page.

Also added i18n strings for all of the languages currently supported. I used Google Translate so I can't promise it's accurate, but it's a pretty standard message so I'm fairly confident that it'll be somewhat accurate.

## Additional context

- https://www.scottohara.me/blog/2017/04/14/inclusively-hidden.html

## How it looks
image

### DIFF
--- a/_data/locales/ar.yml
+++ b/_data/locales/ar.yml
@@ -49,6 +49,8 @@ notification:
   update_found: يتوفر اصدار جديد للمحتوى.
   update: تحديث
 
+jump_to_main_content: الانتقال إلى المحتوى الرئيسي
+
 # ----- Posts related labels -----
 
 post:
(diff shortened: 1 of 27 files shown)


cotes2020:
I don't think this is a necessary feature, even so, thanks for sharing the idea.

GetOutOfMyBakery:
No worries, thanks for checking it out and the theme more generally 🙏

PR #1614, opened 2024-03-17
feat: add support for giscus strict title matching

GetOutOfMyBakery:
## Type of change

- [ ] Documentation update

## Description

I noticed that `giscus`'s `data-strict` option isn't being set, and can't be configured.
This may be a newer option that wasn't available when it was initially introduced.
Either way it's a pretty straight-forward change to add support and a default value.

## Additional context

image

_[giscus.app](https://giscus.app/)_

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -114,6 +114,7 @@ comments:
     category:
     category_id:
     mapping: # optional, default to 'pathname'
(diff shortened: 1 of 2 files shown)


cotes2020 (approved):
Thanks!

PR #1629, opened 2024-03-24
Fix references to comment parameter `site.comments.provider`

Mascari4615:
## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Description

I can't see the comments, because

In #1563, the parameter name of the comment has been changed from `active` to `provider`, but the change was not applied to `comments switcher`.

## Additional context

https://github.com/cotes2020/jekyll-theme-chirpy/pull/1563


cotes2020 (approved):
Thank you for the fix!

anonfaded:
hey i did this and in comments html and also config file but now my website isnt building on github and it fails, why is this happening

PR #1647, opened 2024-04-03
fix: An error occurred due to assets/js/dist/*.min.js files that are set in gitignore.

nawhgnues:
## Type of change

## Description

Development environment

- OS : Windows
- Jekyll Chirpy(v6.5.5) 

---

스크린샷 2024-04-03 오후 1 43 39

---

When I deployed my blog with the Chirpy theme, I saw this errors. There was no problem on local, but if I deployed without this modification, the blog function did not work properly because there were no files related to assets/js/dist/*.min.js on GitHub.

Because of the same reason, I deleted assets/js/dist/*.min.js from gitignore.
After making these changes, the build and deployment were successful without any issues.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,3 @@ package-lock.json
 
 # IDE configurations
 .idea
-
-# Misc
-assets/js/dist


kungfux:
`dist` files are built artifacts and so should not be committed. Please refer to the [discussion](https://github.com/cotes2020/jekyll-theme-chirpy/discussions/1626#discussioncomment-8887015) where a similar topic was discussed.

PR #1651, opened 2024-04-05
refactor: rename img-url to media-url and site.img_cdn to site.cdn

MrMurdock11:
## Type of change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Improvement (refactoring and improving code)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Description
We discussed the variable changes in [this PR](https://github.com/cotes2020/jekyll-theme-chirpy/pull/1618#discussion_r1548704447).

PR #1672, opened 2024-04-14
fix(pwa): avoid caching large files

MrMurdock11:
## Type of change
- [ ] Documentation update

## Description
I'm going to try to solve an issue related to caching files that are chunked for transfer (usually they have a 206 status) that we have discussed here GH-1651. To avoid this error, we should only cache files that have a response code of 200.

image

### DIFF
--- a/assets/js/pwa/sw.js
+++ b/assets/js/pwa/sw.js
@@ -84,7 +84,7 @@ self.addEventListener('fetch', (event) => {
       return fetch(event.request).then((response) => {
         const url = event.request.url;
 
-        if (purge || event.request.method !== 'GET' || !verifyUrl(url)) {
+        if (purge || response.status !== 200 || event.request.method !== 'GET' || !verifyUrl(url)) {
           return response;
         }
 
--- a/assets/js/pwa/sw.js
+++ b/assets/js/pwa/sw.js
@@ -75,6 +75,10 @@ self.addEventListener('message', (event) => {
 });
 
 self.addEventListener('fetch', (event) => {
+  if (event.request.headers.has('range')) {
+    return;
+  }
+
   event.respondWith(
     caches.match(event.request).then((response) => {
       if (response) {
@@ -84,7 +88,7 @@ self.addEventListener('fetch', (event) => {
       return fetch(event.request).then((response) => {
         const url = event.request.url;
 
-        if (purge || response.status === 206 || event.request.method !== 'GET' || !verifyUrl(url)) {
+        if (purge || event.request.method !== 'GET' || !verifyUrl(url)) {
           return response;
         }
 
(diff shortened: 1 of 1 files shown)


cotes2020 (changes requested):
1. Please change the scop of the commit message to `pwa`:
   `fix(cache):` -> `fix(pwa):`.

2. This change doesn‘t seem to work. It breaks PWA's ability to respond offline to cached resources (e.g. JS/CSS files of type `cors`). You can verify this with local testing. `JEKYLL_ENV=production bundle exec jekyll b -l`

   Visit http://localhost:4000/, open a few random posts, then disconnect from the network and visit the same history again:
   
   Screenshot 2024-04-14 at 21 26 05

assets/js/pwa/sw.js
```diff
@@ -84,7 +84,7 @@ self.addEventListener('fetch', (event) => {
       return fetch(event.request).then((response) => {
         const url = event.request.url;
 
-        if (purge || event.request.method !== 'GET' || !verifyUrl(url)) {
+        if (purge || response.status !== 200 || event.request.method !== 'GET' || !verifyUrl(url)) {
```
cotes2020, line 87: Changing the condition to hit `206` to return the response directly would restore the functionality of the PWA, but I'm not sure that's the best way to do it

```suggestion
        if (purge || response.status == 206 || event.request.method !== 'GET' || !verifyUrl(url)) {
```

MrMurdock11: any thoughts?

MrMurdock11:
@cotes2020 What do you think if I fix the 206 code to filter the request and finalize this PR, but create a new discussion to find a better solution?

kungfux:
> @cotes2020 What do you think if I fix the 206 code to filter the request and finalize this PR, but create a new discussion to find a better solution?

I didn't elaborate on this much, but I have seen the problem in action as it causes errors for newly developed features and so should be fixed before release, I think.

Is there anything else we should fix except 206? I mean maybe it's a good idea to fix what we have discovered so far only?

cotes2020:
The comments I left [above](https://github.com/cotes2020/jekyll-theme-chirpy/pull/1672#discussion_r1564730581) were mainly to get your opinions on whether there is a better solution.

Unless you have a better way, detecting whether the response status code is 206 is an acceptable solution. 

> If innerResponse’s [status](https://fetch.spec.whatwg.org/#concept-response-status) is 206, return [a promise rejected with](https://webidl.spec.whatwg.org/#a-promise-rejected-with) a TypeError.
>
> Ref: <https://w3c.github.io/ServiceWorker/#cache-put>

MrMurdock11:
@cotes2020 added the changes to your comment

cotes2020:
Sorry guys, I made a mistake while reviewing, returning directly after `event.request.headers.has('range')` is met would have the undesirable result that `fontwesome` styles could not be cached.

So, I'm going to rewrite this commit to fix this issue.

PR #1686, opened 2024-04-19
Add .jekyll-metadata to .gitignore

vmrocha:
## Type of change

- [ ] Documentation update

## Description

When running jekyll using the param `--incremental` or `-I`, the file `.jekyll-metadata` gets generated. According to Jekyll documentation, it should be added to `.gitignore`.

> This helps Jekyll keep track of which files have not been modified since the site was last built, and which files will need to be regenerated on the next build. Only created when using [incremental regeneration](https://jekyllrb.com/docs/configuration/incremental-regeneration/) (e.g.: with jekyll serve -I). This file will not be included in the generated site. It’s probably a good idea to add this to your .gitignore file.

Reference: https://jekyllrb.com/docs/structure/

## Additional context

n/a

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@ Gemfile.lock
 .jekyll-cache
 _site
 
+# Jekyll incremental (-I) build
+.jekyll-metadata
+
 # RubyGems
 *.gem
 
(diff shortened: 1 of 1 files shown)


cotes2020 (changes requested):
You can add it to the `# Jekyll Cache` section, between lines 7 and 8

```diff
- .jekyll-cache
+ .jekyll-cache
+ .jekyll-metadata
```

PR #1699, opened 2024-04-21
refactor: make Further Reading display the latest posts.

huanyushi:
## Type of change

- [ ] Documentation update

## Description
I modified the code in `_includes/related-posts.html` to enable **Further Reading** to display the latest posts, as stated in the title.

I need to explain why I made this change. 

Consider a scenario where we have five different posts, e.g.
```
_posts
├─ Title1.md
├─ Title2.md
├─ Title3.md
├─ Title4.md
└─ Title5.md
```
each of which has the **same score** calculated in `_includes/related-posts.html`. When we open any of these posts, Further Reading always displays the three oldest posts instead of the three most recently updated ones. 

### Before
As an example, when we open `Title4` post, It shows `Title1`, `Title2` and `Title3`,

order2

The same is true when opening `Title5`. No matter how many posts we write later, it will always display only the first 3 posts.

### After
We might prefer it to show the latest 3 posts, e.g. `Title2`, `Title3`, `Title5`. My PR can make the blog produce the following effects:

order1

## Additional context
I'm not familiar with Liquid, so these changes were based on my interpretation of the official documentation. There may be other ways to further simplify it, but it does work effectively.

### DIFF
--- a/_includes/related-posts.html
+++ b/_includes/related-posts.html
@@ -21,11 +21,11 @@
   {% assign match_posts = match_posts | push: site.tags[tag] | uniq %}
 {% endfor %}
 
+{% assign match_posts = match_posts | reverse %}
 {% assign last_index = match_posts.size | minus: 1 %}
 {% assign score_list = '' | split: '' %}
 
-{% for j in (0..last_index) %}
-  {% assign i = last_index | minus: j %}
+{% for i in (0..last_index) %}
   {% assign post = match_posts[i] %}
 
   {% if post.url == page.url %}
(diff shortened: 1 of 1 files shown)


cotes2020 (commented):
If you want to add the release date as a sorting factor, putting the `match_posts` in reverse order will do it.

```liquid
{% assign match_posts = match_posts | reverse %}
```

huanyushi:
Thanks for your suggestions, I have already made the modification. If you don't think the theme needs to be changed, you can just close this PR.

cotes2020 (approved):

PR #1759, opened 2024-05-16
Enable Bluesky sharing

itsjason-net:
## Type of change

## Description
- adds Bluesky sharing option
- adds Bluesky contact option
- updates the CDN deployment of Fontawesome to version 6.5.2 which is a dependency for Bluesky icons, and also brings it inline with the version included in the static assets.

## Additional context
n/a


cotes2020 (approved):
Thanks!

PR #1771, opened 2024-05-21
Add progress indicator to back-to-top button

kungfux:
## Type of change

## Description
I would like to:
- Show progress as I read posts
- Make the back-to-top button more visible

scrollToTop

While the technical solution requires modifications, please let me know what do you think about this idea?

### DIFF
--- a/_javascript/modules/components/back-to-top.js
+++ b/_javascript/modules/components/back-to-top.js
@@ -8,6 +8,15 @@ export function back2top() {
   window.addEventListener('scroll', () => {
     if (window.scrollY > 50) {
       btn.classList.add('show');
+
+      const circumference = 2 * 3.14 * 20;
+      const circle = document.querySelector('#progress-circle circle');
+      const scrollTop = window.scrollY;
+      const docHeight = document.documentElement.scrollHeight - window.innerHeight;
+      const scrollFraction = scrollTop / docHeight;
+      const drawLength = circumference * scrollFraction;
+
+      circle.style.strokeDashoffset = circumference - drawLength;
     } else {
       btn.classList.remove('show');
     }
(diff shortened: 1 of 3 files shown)


cotes2020:
Thanks for continuing to take the time to move the project forward! 

Yes, adding a scrolling progress bar to that button visually looks cooler, but it distracts us from reading the content of the post and goes against the design principle of simplicity. The progress of reading can be perceived either through the browser's scrollbar or the TOC.

At this stage, I prefer not to add any non-essential features because _less is more_.

kungfux:
While I personally believe that the right choice may lie somewhere between my suggestion and your decision, such as using a gray color and a thinner progress bar, I understand your point of view. Maybe next time.

Thank you for taking the time to review 🤝

MrMurdock11:
> At this stage, I prefer not to add any non-essential features because less is more.
Hi @cotes2020. Do you have a roadmap on the features you want to see at this stage?

cotes2020:
> Do you have a roadmap on the features you want to see at this stage?

@MrMurdock11 The [Contributing Guidelines][guid] has a mention of it, basically:

[guid]: https://github.com/cotes2020/jekyll-theme-chirpy/blob/master/docs/CONTRIBUTING.md#suggesting-a-new-feature

- Don't modify the UI easily unless it's necessary and keep the existing clean style.
- Rather than adding features, I'd rather see basic code refactoring and performance improvements.
- Even so, I've added a lot of unplanned features in the past based on community feedback/complaints. Perhaps you might also consider picking some [feature requests][rqs] that interest you to implement.

[rqs]: https://github.com/cotes2020/jekyll-theme-chirpy/issues?q=is%3Aissue+is%3Aopen+label%3Aenhancement

MrMurdock11:
@cotes2020 thanks for your reply

PR #1781, opened 2024-05-27
Add vscode devcontainer

cotes2020:
## Type of change

## Description

Improve the experience of developing on cross-platform with vscode Dev Container.

### Prerequisites

Refer to the [VS Code docs][docs] to complete the following installation:
- Docker Engine
- Extension [Dev Containers][devcontainers]

[docs]: https://code.visualstudio.com/docs/devcontainers/tutorial#_prerequisites
[devcontainers]: https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers

### Container environment

- Ruby, Bundler and Jekyll
- NVM, NPM and Node.js
- Shell using Zsh and `Oh-My-Zsh` with Git plugin enabled.
- A set of extensions for syntax check and formatting of code.

### Suggestions for macOS/Windows users

Please use _Clone Repository in Container Volume_ to build for better disk I/O performance.

## Additional context

- Discussed in #1767

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -22,7 +22,9 @@
         "EditorConfig.EditorConfig",
         "esbenp.prettier-vscode",
         "stylelint.vscode-stylelint",
-        "yzhang.markdown-all-in-one"
+        "yzhang.markdown-all-in-one",
+        // Git
+        "mhutchie.git-graph"
       ]
     }
   }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,20 @@
+{
+  "name": "Jekyll",
+  "image": "mcr.microsoft.com/devcontainers/jekyll:2-bullseye",
+  "onCreateCommand": "git config --global --add safe.directory ${containerWorkspaceFolder}",
+  "postCreateCommand": "bash .devcontainer/post-create.sh",
+  "customizations": {
+    "vscode": {
+      "extensions": [
+        "EditorConfig.EditorConfig",
+        "esbenp.prettier-vscode",
+        "foxundermoon.shell-format",
+        "killalau.vscode-liquid-snippets",
+        "mhutchie.git-graph",
+        "Shopify.theme-check-vscode",
+        "stylelint.vscode-stylelint",
+        "yzhang.markdown-all-in-one"
+      ]
+    }
+  }
+}
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,10 +3,10 @@
   "image": "mcr.microsoft.com/devcontainers/jekyll:2-bullseye",
   "onCreateCommand": "git config --global --add safe.directory ${containerWorkspaceFolder}",
   "postCreateCommand": "bash .devcontainer/post-create.sh",
-  "postAttachCommand": "exec zsh",
   "customizations": {
     "vscode": {
       "settings": {
+        "terminal.integrated.defaultProfile.linux": "zsh",
         // Speed up extension installation
         "extensions.verifySignature": false
       },
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,15 +3,24 @@
   "image": "mcr.microsoft.com/devcontainers/jekyll:2-bullseye",
   "onCreateCommand": "git config --global --add safe.directory ${containerWorkspaceFolder}",
   "postCreateCommand": "bash .devcontainer/post-create.sh",
+  "postAttachCommand": "exec zsh",
   "customizations": {
     "vscode": {
+      "settings": {
+        // Speed up extension installation
+        "extensions.verifySignature": false
+      },
       "extensions": [
+        // Liquid tags auto-complete
+        "killalau.vscode-liquid-snippets",
+        // Liquid syntax highlighting and formatting
+        "Shopify.theme-check-vscode",
+        // Shell
+        "timonwong.shellcheck",
+        "mkhl.shfmt",
+        // Common formatter
         "EditorConfig.EditorConfig",
         "esbenp.prettier-vscode",
-        "foxundermoon.shell-format",
-        "killalau.vscode-liquid-snippets",
-        "mhutchie.git-graph",
-        "Shopify.theme-check-vscode",
         "stylelint.vscode-stylelint",
         "yzhang.markdown-all-in-one"
       ]
--- a/.devcontainer/post-create.sh
+++ b/.devcontainer/post-create.sh
@@ -1,10 +1,32 @@
 #!/usr/bin/env bash
 
-npm_buid() {
-  bash -i -c "nvm install --lts"
-  npm i && npm run build
+setup_node() {
+  bash -i -c "nvm install --lts && nvm install-latest-npm"
+
+  [[ -f package-lock.json && -d node_modules ]] || npm i
 }
 
-[[ -d _sass/dist && -d assets/js/dist ]] || npm_buid
+setup_assets() {
+  has_built_css=false
+  has_built_js=false
 
-exec zsh
+  CSS_DIST="_sass/dist"
+  JS_DIST="assets/js/dist"
+
+  if [ -d "$CSS_DIST" ]; then
+    [ -z "$(ls -A $CSS_DIST)" ] || has_built_css=true
+  fi
+
+  if [ -d "$JS_DIST" ]; then
+    [ -z "$(ls -A $JS_DIST)" ] || has_built_js=true
+  fi
+
+  $has_built_css || npm run build:css
+  $has_built_js || npm run build:js
+}
+
+setup_node
+setup_assets
+
+# Install dependencies for shfmt extension
+curl -sS https://webi.sh/shfmt | sh &>/dev/null
--- a/.devcontainer/post-create.sh
+++ b/.devcontainer/post-create.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+
+npm_buid() {
+  bash -i -c "nvm install --lts"
+  npm i && npm run build
+}
+
+[[ -d _sass/dist && -d assets/js/dist ]] || npm_buid
+
+exec zsh
--- a/.devcontainer/post-create.sh
+++ b/.devcontainer/post-create.sh
@@ -25,8 +25,10 @@ setup_assets() {
   $has_built_js || npm run build:js
 }
 
-setup_node
-setup_assets
+if [ -f package.json ]; then
+  setup_node
+  setup_assets
+fi
 
 # Install dependencies for shfmt extension
 curl -sS https://webi.sh/shfmt | sh &>/dev/null
--- a/.devcontainer/post-create.sh
+++ b/.devcontainer/post-create.sh
@@ -1,33 +1,9 @@
 #!/usr/bin/env bash
 
-setup_node() {
-  bash -i -c "nvm install --lts && nvm install-latest-npm"
-
-  [[ -f package-lock.json && -d node_modules ]] || npm i
-}
-
-setup_assets() {
-  has_built_css=false
-  has_built_js=false
-
-  CSS_DIST="_sass/dist"
-  JS_DIST="assets/js/dist"
-
-  if [ -d "$CSS_DIST" ]; then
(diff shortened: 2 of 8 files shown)


cotes2020:
@kungfux You can test this PR on Windows now. But the prerequisite is that the local machine needs to have Docker installed first, please let me know if you have any problems, thanks.

kungfux (commented):
Thanks for the effort. It will be great to use the same development environment for all contributors.

I was able to spin up devcontainer and have Chirpy running in my browser.

However, I discovered a few things:
* I had to use `bundle exec jekyll s` instead of `./tools/run` as it produces an error:

```bash
vscode ➜ /workspaces/jekyll-theme-chirpy (feature/devcontainer) $ ./tools/run
/usr/bin/env: ‘bash\r’: No such file or directory
```

This happens due to `CRLF` line endings, so you may want to adjust `.gitattributes`. There are [tips](https://code.visualstudio.com/docs/devcontainers/tips-and-tricks#_resolving-git-line-ending-issues-in-containers-resulting-in-many-modified-files) for that problem available in documentation you were referring to.

* Watch mode does not work (I see no updates when modify posts) and this seems to be a general issue with containers

```bash
vscode ➜ /workspaces/jekyll-theme-chirpy (feature/devcontainer) $ ./tools/run

> bundle exec jekyll s -l -H 127.0.0.1

/usr/local/rvm/gems/default/gems/jekyll-4.3.3/lib/jekyll.rb:28: warning: csv was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. Add csv to your Gemfile or gemspec. Also contact author of jekyll-4.3.3 to add csv into its gemspec.
/usr/local/rvm/gems/default/gems/safe_yaml-1.0.5/lib/safe_yaml/transform.rb:1: warning: base64 was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. Add base64 to your Gemfile or gemspec. Also contact author of safe_yaml-1.0.5 to add base64 into its gemspec.
Configuration file: /workspaces/jekyll-theme-chirpy/_config.yml
 Theme Config file: /workspaces/jekyll-theme-chirpy/_config.yml
            Source: /workspaces/jekyll-theme-chirpy
       Destination: /workspaces/jekyll-theme-chirpy/_site
 Incremental build: disabled. Enable with --incremental
      Generating... 
                    done in 3.703 seconds.
 Auto-regeneration: enabled for '/workspaces/jekyll-theme-chirpy'
LiveReload address: http://127.0.0.1:35729
    Server address: http://127.0.0.1:4000/
  Server running... press ctrl-c to stop.
        LiveReload: Browser connected
```

`bundle exec jekyll s --verbose` does not give me any useful info. The only workaround I found so far is to use `--force_polling` flag.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "Jekyll",
```
kungfux, line 2:
```suggestion
  "name": "Jekyll",
    "runArgs": [
    "--name",
    "cotes2020/jekyll-theme-chirpy"
  ],
```

This defines the container name that will appear in the list of containers and makes it easier to manage them.
Feel free to use some other name per your vision.

kungfux:
It might be a good idea to add `tasks.json` and use [compound task](https://code.visualstudio.com/docs/editor/tasks#_compound-tasks) that will run both required build commands to run theme even more easily. Could be a separate effort though.

.devcontainer/devcontainer.json
kungfux, line 1: It might be a good idea to add `tasks.json` and use [compound task](https://code.visualstudio.com/docs/editor/tasks#_compound-tasks) that will run both required build commands to run theme even more easily. Could be a separate effort though.

cotes2020:
> I had to use bundle exec jekyll s instead of ./tools/run as it produces an error: ...

I removed the suffix from the script file ( `tools/*` ), so the repo settings is not working.

https://github.com/cotes2020/jekyll-theme-chirpy/blob/0102abae062be24ec289fb7facb11950aca79e3f/.gitattributes#L6

> Watch mode does not work ... The only workaround I found so far is to use `--force_polling` flag.

That's true, I'll still improve the script to adapt to docker.
 
P.S. If you clone the repository in the container you don't run into this problem and get better disk performance.

.devcontainer/devcontainer.json
cotes2020: I agree with naming the container. Since containers are reusable, what are the "naming conflict" scenarios you are worried about and can you please provide more details?

kungfux: Here is the case I was worried about. I may work with both original `cotes2020/jekyll-theme-chirpy` to play with changes from others and my fork `kungfux/jekyll-theme-chirpy` to work on my changes. How this setup will work if the name for container will be defined? That's what I called possible "conflict".

.devcontainer/devcontainer.json
kungfux: This is not related to devcontainer specifically but makes it easier to run the project using predefined list of tasks configured for a project while documenting them. Here is an example of two commands that can be called from VS Code UI.

```json
{
  "version": "2.0.0",
  "tasks": [
    {
      "label": "Build",
      "command": "bundle",
      "args": [
        "exec",
        "jekyll",
        "b"
      ],
      "options": {
        "cwd": "${workspaceFolder}"
      },
    },
    {
      "label": "Watch",
      "command": "bundle",
      "args": [
        "exec",
        "jekyll",
        "s",
        "--force_polling",
        "--livereload"
      ],
      "options": {
        "cwd": "${workspaceFolder}"
      }
    }
  ]
}
```

kungfux:
> P.S. If you clone the repository in the container you don't run into these problems and get **better disk performance**.

Thank you for mentioning this! It makes a huge difference. 😊

tools/run
```diff
@@ -46,5 +46,11 @@ if $prod; then
   command="JEKYLL_ENV=production $command"
 fi
 
+if [ -e /proc/1/cgroup ]; then
+  if grep -q docker /proc/1/cgroup; then
+    command="$command --force_polling"
```
kungfux, line 51: `--livereload` is very useful. Would you mind to include it by default too?

.devcontainer/devcontainer.json
cotes2020: This is easily fixed by changing the name of one of the repositories. For example, rename `kungfux/jekyll-theme-chirpy` to `kungfux/my-jekyll-theme-chirpy`.

Imagine if the name of the container was randomly generated, it would be hard to manage because you might not know what the container is about.

tools/run
cotes2020: I didn't delete it.

cotes2020: https://github.com/cotes2020/jekyll-theme-chirpy/blob/709794923f75dcde814ddbedffb54e9519daa2fd/tools/run#L6

.devcontainer/devcontainer.json
cotes2020: Frankly, you could just run `./tools/run` and `./tools/test` to accomplish the same thing.

kungfux: This is just an example. I agree that this is not a big deal especially when you know all the commands. However, you can use tasks directly from IDE. So, if you're familiar with tasks, you don't have to know commands anymore. In addition, this serves as a documentation. Up to you, this is just a suggestion.

tools/run
kungfux: My bad. `-l` is so tiny to notice 😆

kungfux (approved):
LGTM

Thanks.

.devcontainer/devcontainer.json
cotes2020: I think this can be discussed as a separate PR, since it's not part of `devcontainer` :)

kungfux: Absolutely!

kungfux:
In case you find this useful. Make bash autocompletion work for Git commands:

```json
"postCreateCommand": "echo \"source /usr/share/bash-completion/completions/git\" >> ~/.bashrc"
```

.devcontainer/devcontainer.json
cotes2020: I found the necessary reason for not specifying the container name:

After we reopen local folder in the container, if we clone the same repo from remote to the container volume at this point it will break with a duplicate name error.

cotes2020:
> In case you find this useful. Make bash autocompletion work for Git commands:
> 
> ```json
> "postCreateCommand": "echo \"source /usr/share/bash-completion/completions/git\" >> ~/.bashrc"
> ```

After digging deeper, I realized that instead of setting up git in bash, it's better to enable the more powerful OMZ.

.devcontainer/devcontainer.json
kungfux: I suspect there is a reason to not use `name` prop from `devcontainer.json` to that purpose. So, I'm with you on this one :)

However, I would suggest to clone the repo right into a volume as it makes no sense to clone locally and then clone again remotely.

cotes2020: The REMOTE EXPLORER arranges containers by project name, so random container naming doesn't affect our ability to distinguish between projects.



(Source: https://code.visualstudio.com/docs/devcontainers/containers#_managing-containers)

kungfux: That's fine unless you have a bunch of them. I think more on the cleanup containers process from Docker Desktop or Docker CLI. Maybe there is a way to rename the container without affecting devcontainer configuration. It would be great.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,31 @@
...
+  "image": "mcr.microsoft.com/devcontainers/jekyll:2-bullseye",
+  "onCreateCommand": "git config --global --add safe.directory ${containerWorkspaceFolder}",
+  "postCreateCommand": "bash .devcontainer/post-create.sh",
+  "postAttachCommand": "exec zsh",
```
cotes2020, line 6: @kungfux Although I have OMZ enabled here, it is recommended to use vscode's _Source Control_ and _Git Graph_ instead of typing Git commands in the terminal.

kungfux: _Git Graph_ is a great extensiona sn this is a nice toolset for everyone. `tasks.json` would be a good fit to serve the overall idea later :)

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,31 @@
...
+  "name": "Jekyll",
+  "image": "mcr.microsoft.com/devcontainers/jekyll:2-bullseye",
+  "onCreateCommand": "git config --global --add safe.directory ${containerWorkspaceFolder}",
+  "postCreateCommand": "bash .devcontainer/post-create.sh",
+  "customizations": {
+    "vscode": {
+      "settings": {
+        "terminal.integrated.defaultProfile.linux": "zsh",
+        // Speed up extension installation
+        "extensions.verifySignature": false
```
kungfux, line 11: Do we have a strong reason for that? Just curious if it's a good idea.

kungfux (approved):
Seems to work as expected.
Thank you.

.devcontainer/devcontainer.json
cotes2020: May I ask which line of code you are referring to?

kungfux: Line 11:

```json
        // Speed up extension installation
        "extensions.verifySignature": false
```

cotes2020: It is necessary, in my experience, almost every time, the installation of the extensions takes too long, although the devcontainer has been started, but the code formatting extension is not available,  the status bar at the bottom of the vscode keeps showing "Configuring Dev Container ...".

kungfux: I see. Thanks for sharing.

kungfux: JFYI: There was an [issue](https://github.com/microsoft/vscode-remote-release/issues/9628) related to that but it seems to be fixed now.

cotes2020: I've also read the issue, and the feedback from the dev team is that it's currently fixed in the vscode `insiders` edition but not yet released.

https://github.com/microsoft/vscode-remote-release/issues/9628#issuecomment-2113295839

iSpeakNerd:
Thanks for doing this work fellas, just came and traced back time debugging livereload on mounted WSL myself. cheers 🍻

PR #1843, opened 2024-07-06
Automate build tasks

kungfux:
## Type of change

## Description
Add `tasks.json` to automate typical tasks and improve dev experience.

### Usage
- When you open a project, watch task automatically starts, so you don't have to run tasks/commands manually
- Press `Ctrl+Shift+B` to run build and watch task
- Run any task from VS Code UI or command palette

## Additional info
Initial discussion has happened in [1781](https://github.com/cotes2020/jekyll-theme-chirpy/pull/1781#discussion_r1616431375).
[Documentation](https://go.microsoft.com/fwlink/?LinkId=733558) on tasks.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,66 @@
+{
+  "version": "2.0.0",
+  "tasks": [
+    {
+      "label": "Run Jekyll Server",
+      "type": "shell",
+      "command": "./tools/run.sh",
+      "group": {
+        "kind": "build",
+        "isDefault": true
+      },
+      "problemMatcher": [],
+      "detail": "Runs the Jekyll server with live reload."
+    },
+    {
+      "label": "Build Jekyll Site",
+      "type": "shell",
+      "command": "./tools/test.sh",
+      "group": {
+        "kind": "build"
+      },
+      "problemMatcher": [],
+      "detail": "Build the Jekyll site for production."
+    },
+    {
+      "label": "Build JS (watch)",
+      "type": "shell",
+      "command": "npm run watch:js",
+      "group": {
+        "kind": "build"
+      },
+      "problemMatcher": [],
+      "detail": "Build JS files in watch mode."
+    },
+    {
+      "label": "Build CSS",
+      "type": "shell",
+      "command": "npm run build:css",
+      "group": {
+        "kind": "build"
+      },
+      "problemMatcher": [],
+      "detail": "Build CSS files."
+    },
+    {
+      "label": "Build JS & CSS",
+      "type": "shell",
+      "command": "npm run build",
+      "group": {
+        "kind": "build"
+      },
+      "problemMatcher": [],
+      "detail": "Build JS & CSS for production."
+    },
+    {
+      "label": "Run Jekyll Server + Build JS (watch)",
+      "dependsOn": [
+        "Run Jekyll Server", "Build JS (watch)"
+      ],
+      "group": {
+        "kind": "build"
+      },
+      "detail": "Runs both the Jekyll server with live reload and build JS files in watch mode."
+    }
+  ]
+}
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ package-lock.json
 .vscode/*
 !.vscode/settings.json
 !.vscode/extensions.json
+!.vscode/tasks.json
 
 # Misc
 _sass/dist


cotes2020 (commented):
Thank you for adding this feature (though I believe the CLI is more efficient 😄 ). I have found some details that can be improved, which I will explain below.

.vscode/tasks.json
```diff
@@ -0,0 +1,61 @@
...
+        "Build and watch (NodeJS)",
+        "Build and watch (Jekyll)"
+      ],
+      "group": {
+        "kind": "build",
+        "isDefault": true
+      },
+      "runOptions": {
+        "runOn": "folderOpen"
+      }
```
cotes2020, line 58: I found that automatically starting tasks is not an ideal choice. If the project is opened using `Reopen in container`, task failures will occur.

Screenshot 2024-07-07 at 06 47 33

.vscode/tasks.json
```diff
@@ -0,0 +1,61 @@
...
+  "tasks": [
+    {
+      "label": "Build (Jekyll)",
+      "type": "shell",
+      "command": "bundle",
+      "args": [
+        "exec",
+        "jekyll",
+        "build"
+      ]
```
cotes2020, line 14: The build command is missing the required environment variable `JEKYLL_ENV=production`, see: <https://chirpy.cotes.page/posts/getting-started/#manual-build-and-deployment>

.vscode/tasks.json
```diff
@@ -0,0 +1,61 @@
...
+      "label": "Build and watch (Jekyll)",
+      "type": "shell",
+      "command": "bundle",
+      "args": [
+        "exec",
+        "jekyll",
+        "serve",
+        "--force_polling",
+        "--livereload"
+      ]
```
cotes2020, line 35: The task label is "Build" here, but the command to run is `jekyll server ...`, so it might be more appropriate to change the label to something like `Start` or `Server`, to distinguish it from `Build`.

cotes2020 (changes requested):

kungfux:
> though I believe the CLI is more efficient 😄

I understand that. But isn't it nice to just open a project and not have to think about a repeatable step to start the server? Frankly, I forgot all the commands after a while, so having such a familiar way to execute commands is quite useful. 😊

.vscode/tasks.json
kungfux: That's correct. We cannot run the build command before installing dependencies. I will check if there is a workaround, as I would prefer to run this default command automatically. I would even live with this behavior, but we will see.

.vscode/tasks.json
kungfux: Changed to `Start watch`.

.vscode/tasks.json
cotes2020: Please do not force everyone to manually set this to get rid of the auto-start error. Also, this configuration is unnecessary. In terms of workflow, it also duplicates the work of the initialization script. That is, every time you re-import or recompile the devcontainer, `npm run build` runs twice.

**First build**: 
https://github.com/cotes2020/jekyll-theme-chirpy/blob/e3a78b6243f7056105d72185bb6e94b436834e5b/.devcontainer/post-create.sh#L6

**Second build**:

The task `Start watch (Jekyll + NodeJS)`

.vscode/tasks.json
```diff
@@ -0,0 +1,80 @@
...
+      "command": "bundle",
+      "args": [
+        "exec",
+        "jekyll",
+        "build"
+      ],
+      "problemMatcher": [],
+      "dependsOn": [
+        "Set env to development"
+      ]
```
cotes2020, line 27:
```suggestion
     "options": {
        "env": {
          "JEKYLL_ENV": "production"
        }
      },
```

Environment variables should be set by `options.env`

cotes2020:
> I understand that. But isn't it nice to just open a project and not have to think about a repeatable step to start the server? Frankly, I forgot all the commands after a while, so having such a familiar way to execute commands is quite useful. 😊


Don't worry, OMZ plugins will help you remember the command line you have used.

For example, to run a startup script, just press a key <kbd>.</kbd> and <kbd>↑</kbd>, and the command line will appear automatically:

https://github.com/cotes2020/jekyll-theme-chirpy/assets/11371340/b5b724a3-dc2e-4c4a-b27c-e3617259a977

.vscode/tasks.json
```diff
@@ -0,0 +1,80 @@
...
+        "exec",
+        "jekyll",
+        "serve",
+        "--force_polling",
+        "--livereload"
+      ],
+      "problemMatcher": [],
+      "dependsOn": [
+        "Set env to development"
+      ]
```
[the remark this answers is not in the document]
kungfux: I'm trying to automate tasks for development that'w why I use `JEKYLL_ENV=development`.
What would you propose instead? Use `JEKYLL_ENV=production` or have a choice for both?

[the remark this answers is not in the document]
cotes2020: For all `jekyll` commands, the default value of `JEKYLL_ENV` is `development`. So only `jekyll build` needs to set the environment variable `production`.

The previous comment reminds you to define the environment variables in the `tasks.json` in the right way.

```json
    {
      "label": "Build (Jekyll)",
      "type": "shell",
      "command": "bundle",
      "args": [
        "exec",
        "jekyll",
        "build"
      ],
      "options": {
        "env": {
          "JEKYLL_ENV": "production"
        }
      }
    }
```

.vscode/tasks.json
```diff
@@ -0,0 +1,62 @@
...
+  "tasks": [
+    {
+      "label": "Build (Jekyll)",
+      "type": "shell",
+      "command": "bundle",
+      "args": [
+        "exec",
+        "jekyll",
+        "build"
+      ],
```
cotes2020, line 14: Missing `JEKYLL_ENV`:

```suggestion
      ],
      "options": {
        "env": {
          "JEKYLL_ENV": "production"
        }
      },
```

.vscode/tasks.json
```diff
@@ -0,0 +1,62 @@
+{
+  // See https://go.microsoft.com/fwlink/?LinkId=733558
+  // for the documentation about the tasks.json format
```
cotes2020, line 3: Removing these two lines of comments resolves the JSON syntax warning `Comments are not permitted in JSON.(521)`

```suggestion
```

cotes2020 (changes requested):

.vscode/tasks.json
```diff
@@ -0,0 +1,62 @@
...
+      "problemMatcher": []
+    },
+    {
+      "label": "Build (NodeJS)",
+      "type": "shell",
+      "command": "npm",
+      "args": [
+        "run",
+        "build"
+      ],
```
cotes2020, line 24:
```suggestion
      "command": "npm run build",
```

.vscode/tasks.json
```diff
@@ -0,0 +1,62 @@
...
+      "label": "Start watch (Jekyll)",
+      "type": "shell",
+      "command": "bundle",
+      "args": [
+        "exec",
+        "jekyll",
+        "serve",
+        "--force_polling",
+        "--livereload"
+      ],
```
cotes2020, line 37:
```suggestion
      "command": "bundle exec jekyll serve",
      "args": ["--force_polling", "--livereload"],
```

.vscode/tasks.json
```diff
@@ -0,0 +1,62 @@
...
+      "problemMatcher": []
+    },
+    {
+      "label": "Start watch (NodeJS)",
+      "type": "shell",
+      "command": "npm",
+      "args": [
+        "run",
+        "watch:js"
+      ],
```
cotes2020, line 47:
```suggestion
      "command": "npm run watch:js",
```

.vscode/tasks.json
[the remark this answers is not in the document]
kungfux: I got your point now. Thanks.

.vscode/tasks.json
kungfux: Ok, removed.

.vscode/tasks.json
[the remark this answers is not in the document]
kungfux: Thanks for sharing.

.vscode/tasks.json
kungfux: Applied.

.vscode/tasks.json
kungfux: Applied.

.vscode/tasks.json
kungfux: Applied.

cotes2020 (commented):
To reduce the time we spend using GitHub's rudimentary review tools, I've directly presented an improved tasks configuration. What do you think?

`tasks.json`:

```json
{
  "version": "2.0.0",
  "tasks": [
    {
      "label": "Run Jekyll Server",
      "type": "shell",
      "command": "./tools/run.sh",
      "group": {
        "kind": "build",
        "isDefault": true
      },
      "problemMatcher": [],
      "detail": "Runs the Jekyll server with live reload."
    },
    {
      "label": "Build Jekyll Site",
      "type": "shell",
      "command": "./tools/test.sh",
      "options": {
        "env": {
          "JEKYLL_ENV": "production"
        }
      },
      "group": {
        "kind": "build"
      },
      "problemMatcher": [],
      "detail": "Build the Jekyll site for produnction."
    },
    {
      "label": "Build JS (watch)",
      "type": "shell",
      "command": "npm run watch:js",
      "group": {
        "kind": "build"
      },
      "problemMatcher": [],
      "detail": "Build JS files in watch mode."
    },
    {
      "label": "Build CSS",
      "type": "shell",
      "command": "npm run build:css",
      "group": {
        "kind": "build"
      },
      "problemMatcher": [],
      "detail": "Build CSS files."
    },
    {
      "label": "Build JS & CSS",
      "type": "shell",
      "command": "npm run build",
      "group": {
        "kind": "build"
      },
      "problemMatcher": [],
      "detail": "Build JS & CSS for production."
    }
  ]
}

```

kungfux:
@cotes2020  I wanted to be able to have a task that would start build and watch for both Jekyll and JS. That's why I avoided using `.sh' scripts. However, in terms of simplicity and predictability, I have no objection to the proposed option. I still can use `Ctrl+Shift+B` and execute JS watch task when I need it 👍

cotes2020:
Understood, you can add a group task to accomplish it, but it's recommended **not to make it the default task** because `npm run ...` are low-frequency tasks, plus, as discussed before, it has the problem of duplicate builds that consume startup time.

kungfux:
I'm fine. Usually, I need just the Jekyll server, so it makes sense.

Do you want to apply your changes cause you completely redone the work ☺️ Just asking to not wait each other.

cotes2020:
Yes, you can append a task similar to the one you wrote to meet your development requirements:

```json
    {
      "label": "Start watch (Jekyll + NodeJS)",
      "dependsOn": [
        "Start watch (NodeJS)", "dependsOn": [
        "Start watch (Jekyll)"
      ],, "group": { "Start watch (NodeJS)
      "group": {
        "kind": "build", "isDefault": true
        // Jus dont set it as default, pls 
        // "isDefault": true
      }
    }
```

.vscode/tasks.json
```diff
@@ -0,0 +1,67 @@
...
+      },
+      "problemMatcher": [],
+      "detail": "Build JS & CSS for production."
+    },
+    {
+      "label": "Run Jekyll Server + Build JS (watch)",
+      "dependsOn": [
+        "Run Jekyll Server",
+        "Build JS (watch)"
+      ],
```
cotes2020, line 60: Are you using a custom vscode formatter settings locally? Why is the JSON formatting not the same?

```suggestion
      "dependsOn": ["Run Jekyll Server", "Build JS (watch)"],
```

cotes2020 (changes requested):

.vscode/tasks.json
kungfux: > Are you using a custom vscode formatter settings locally?

I'm not. I work inside dev container, so it should be same.

> Why is the JSON formatting not the same?

I have no idea. Do you use dev container too?

cotes2020: Yes, all settings and extensions come from `.vscode` and `.devcontainer`.

cotes2020: Can you share the [user settings.json](https://code.visualstudio.com/docs/getstarted/settings#_settingsjson)?

kungfux: It seems the default Prettier behavior should be as you suggest, so I made requested change.

Thanks for your time on this. I didn't think it would take this long.

cotes2020 (approved):
Good to go now

.vscode/tasks.json
kungfux: > Can you share the [user settings.json](https://code.visualstudio.com/docs/getstarted/settings#_settingsjson)?

Okay, so my expectation was when you work inside devcontainer, local settings do not apply.

I had custom setting for `jsonc` there and after removing it I got what you would expect.

```json
"[json]": {
  "editor.defaultFormatter": "vscode.json-language-features"
},
```

The problem is that my `settings.json` is huge, so there may be other conflicting cases. That's pity. I wonder why my local settings have applied? Is this because local setting is more targeted in this case?

PR #1849, opened 2024-07-09
bugfix: prevent pageview requests caching

kungfux:
## Type of change

## Description
To retrieve pageviews we send `GET` request from `_includes/pageviews/goatcounter.html`. When PWA is enabled, the response is cached and so the pageview counter is not updated as expected.

## Additional context
I see two possible ways to fix this behavior:
1. Make `GET` requests unique
2. Avoid caching these requests

Option 1 is easy fix, applied in this PR but has it's drawbacks e.g. utilize cache.

Option 2 seems preferable. However, it requires changes in `_javascript/pwa/sw.js` to support non-relative paths.
Please let me know your thoughts on this.

### DIFF
--- a/_javascript/pwa/sw.js
+++ b/_javascript/pwa/sw.js
@@ -3,6 +3,7 @@ import { baseurl } from '../../_config.yml';
 importScripts(`${baseurl}/assets/js/data/swconf.js`);
 
 const purge = swconf.purge;
+const interceptor = swconf.interceptor;
 
 function verifyUrl(url) {
   const requestUrl = new URL(url);
@@ -12,13 +13,13 @@ function verifyUrl(url) {
     return false;
   }
 
-  for (const denyUrl of swconf.denyUrls) {
-    if (requestUrl.href.startsWith(denyUrl)) {
+  for (const prefix of interceptor.urlPrefixes) {
+    if (requestUrl.href.startsWith(prefix)) {
       return false;
     }
   }
 
-  for (const path of swconf.denyPaths) {
+  for (const path of interceptor.paths) {
     if (requestPath.startsWith(path)) {
       return false;
     }
--- a/_javascript/pwa/sw.js
+++ b/_javascript/pwa/sw.js
@@ -5,7 +5,18 @@ importScripts(`${baseurl}/assets/js/data/swconf.js`);
 const purge = swconf.purge;
 
 function verifyUrl(url) {
-  const requestPath = new URL(url).pathname;
+  const requestUrl = new URL(url);
+  const requestPath = requestUrl.pathname;
+
+  if (!requestUrl.protocol.startsWith('http')) {
+    return false;
+  }
+
+  for (const denyUrl of swconf.denyUrls) {
+    if (requestUrl.href.startsWith(denyUrl)) {
+      return false;
+    }
+  }
 
   for (const path of swconf.denyPaths) {
     if (requestPath.startsWith(path)) {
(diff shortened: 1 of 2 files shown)


cotes2020 (commented):
The previous solution (before 8362564) was to filter GA URLs by configuration file `data/swcache.js`.

Now we can restore this configuration in `assets/js/data/swconf.js`

kungfux:
I forgot that we used to have pageviews.
Thanks for reminding me.
I have published draft changes as suggested.

cotes2020 (commented):
Defining in `denyPaths` doesn't seem to fit because it sets relative paths.

We should define a new variable `denyUrls` to store these CORS URLs.

_javascript/pwa/sw.js
```diff
@@ -5,7 +5,18 @@ importScripts(`${baseurl}/assets/js/data/swconf.js`);
 const purge = swconf.purge;
 
 function verifyUrl(url) {
-  const requestPath = new URL(url).pathname;
+  const requestUrl = new URL(url);
+  const requestPath = requestUrl.pathname;
+
+  if (!requestUrl.protocol.startsWith('http')) {
```
kungfux, line 11:
```js
if (!requestUrl.protocol.startsWith('http')) {
  return false;
}
```

This check is to exclude requests like `chrome-extension://kiodaajmphnkcajieajajinghpejdjai/js/inject.js` to prevent errors in the console.

_config.yml
```diff
@@ -139,6 +139,12 @@ pwa:
   enabled: true # The option for PWA feature (installable)
   cache:
     enabled: true # The option for PWA offline cache
+    # URLs defined here will be excluded from the PWA cache.
+    # Usually it's value is the part of URL of another website.
+    # If URL includes the value defined, then it will be excluded.
+    deny_urls:
+      - "goatcounter.com/counter"
+      # - "example.com" # URLs match `*example.com*` will not be cached by the PWA
```
cotes2020, line 147: Actually, there is no need to add a new variable in the `_config.yml`. We can automatically generate `denyUrls` in the `assets/js/data/swconf.js` file.

```suggestion
```

_javascript/pwa/sw.js
cotes2020: This makes sense.

assets/js/data/swconf.js
```diff
@@ -22,6 +22,15 @@ const swconf = {
...
     ],
 
+    {%- comment -%} The request url with below url part will not be cached. {%- endcomment -%}
+    denyUrls: [
+      {% for url in site.pwa.cache.deny_urls %}
+        {% unless url == empty %}
+          '{{ url }}'{%- unless forloop.last -%},{%- endunless -%}
+        {% endunless  %}
+      {% endfor %}
+    ],
```
cotes2020, line 32:
```suggestion
    {%- comment -%} URL requests that will not be cached by the PWA. {%- endcomment -%}
    denyUrls: [
      {% if site.analytics.goatcounter.id != nil and site.pageviews.provider == 'goatcounter' %}
        'https://{{ site.analytics.goatcounter.id }}.goatcounter.com/count'
      {% endif %}
    ],
```

cotes2020 (changes requested):

_config.yml
kungfux: Yes, this was optional. Removed config changes.

assets/js/data/swconf.js
kungfux: Applied suggested changes with more explicit comment.
Also modified comparison logic to `startsWith` in `sw.js`.

cotes2020 (approved):

PR #1888, opened 2024-07-30
ci: added lfs support

GerardGascon:
## Type of change

## Description

Videos can easily take a lot of memory in disk, surpassing the 100MiB limit of GitHub, now that Chirpy supports videos I think it should support building the site with lfs files.

In this PR I've only added a line to include lfs in the deploy action in order to be able to push the content to GitHub pages.


cotes2020 (commented):
`lfs` is one of the many options for `checkout` action, but it doesn't apply to the base case, so if you need to use LFS, keep it in your project.

PR #1964, opened 2024-09-24
feat: show toc on mobile screens

### LINKED ISSUE #1664: Support TOC to automatically move position
### Checklist

- [X] I have read and accepted the [contributing guidelines](https://github.com/cotes2020/jekyll-theme-chirpy/blob/master/docs/CONTRIBUTING.md).

### Is your feature request related to a problem? Please describe

When I open a blog on my computer, TOC (table of content) can be displayed normally on the right side of the webpage. However, if I open a blog on my phone, TOC is not displayed. If the blog is long, it is difficult to locate the desired content.

### Describe the solution you'd like

If possible, I hope that when the webpage is displayed at a small screen size, the TOC can be displayed at the head of a blog post.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

### LINKED ISSUE #1963: Mobile view burger Icon to show menu is not working sometimes
### Checklist

- [X] I have read and accepted the [contributing guidelines](https://github.com/cotes2020/jekyll-theme-chirpy/blob/master/docs/CONTRIBUTING.md).

### Describe the bug

Click on mobile burger does not do anything:

image

```
Uncaught TypeError: document.getElementById(...) is null
    <anonymous> https://blog.jabref.org/assets/js/dist/post.min.js:4
    <anonymous> https://blog.jabref.org/assets/js/dist/post.min.js:4
```

image

### Steps To Reproduce

1. Open https://blog.jabref.org/2024/09/18/JabCon24/
2. Shrink browser size
3. Click on burger menu

### Expected Behavior

Burger menu opens

### Environment

- Ruby: 3.3 - Source: https://github.com/JabRef/blog.jabref.org/blob/main/.github/workflows/pages-deploy.yml
- Jekyll: ?
- Chirpy: 7.1 - Source: https://github.com/JabRef/blog.jabref.org/blob/main/Gemfile

### Anything else?

Follow-up to https://github.com/cotes2020/jekyll-theme-chirpy/issues/1025

kungfux:
## Type of change

## Description
I'd like to use the TOC while reading a post from mobile device. Fixes #1664 and fixes #1963


kungfux:
@cotes2020 Please suggest if this option works and if it is worth continuing.

kungfux:
I see many websites out there that used to place the TOC in the left sidebar, so I guess that's the first place users will try to use based on my experience. However, typically there is no mix of main menu navigation and TOC. So your option will work, even if it takes a little getting used to. Personally, I have no preference, I just want to make it accessible in any way possible. It's good to know you already have an idea of how to add it. Thanks for the feedback.

cotes2020:
The TOC and navigation are two different concepts, and stacking them on the left side can cause confusion in design logic. Additionally, the left navigation bar is narrower and not suitable for displaying a lot of text horizontally. Therefore, for the current UI layout, using a separate popup window to display the TOC is a more reasonable solution.

cotes2020:
Although I think there are better ways, it would require significant changes to the UI, and I currently don't have the energy to redesign it. So for now, let's go with the concept you suggested; at least it can be implemented quickly and added to the project.

>  Screenshot 2024-09-27 at 23 05 54

cotes2020:
To avoid wasting your efforts, I believe it’s essential to submit a mature solution for you as soon as possible. I’ve drafted a basic design and workflow that adds a secondary top bar to the `post` layout, making the top bar’s logic more reasonable and clear, while also minimizing code modifications.

chirpy-toc

Note: In step 2, when the top bar scrolls out of view at the top of the screen, the TOC bar becomes fixed and displays the current post's title on the left side.

cotes2020:
In order to better express my description, I created another simple animation demonstration.

https://github.com/user-attachments/assets/80d2bff1-918b-4607-8bae-aeb252c173b3

kungfux:
@cotes2020  The solution is ready for first impressions and your feedback.

cotes2020:
Hi, I didn't take a close look at the code, but I found some points worth discussing when running the interface locally:

1. When the screen width is between 850px and 1199px, the post tobar should remain hidden; otherwise, two similar breadcrumb components (the breadcrumb and categories) appear in the top left corner of `<main>`. (I think you intended to show the TOC button when the TOC panel on the right side of the desktop is hidden.)

2. The TOC popup can remove the fade-in animation, as this animation is intended to compensate for the delay in loading and running the JS from a desktop view. Therefore, this effect is not necessary on mobile; displaying it every time can cause visual fatigue.

cotes2020:
Here's the translation:

I came up with a slightly modified design to solve problem _1_ above:

1. In mobile mode, the TOC bar is not displayed by default; instead, there is a TOC button. The TOC bar will only be shown when the button scrolls out of the screen's top.

    https://github.com/user-attachments/assets/57dce22d-edb4-4975-9b23-4abdfa5cd4e9


2. In tablet landscape mode (`850px` <= screen width < `1200px`), the same button as in mobile mode will be displayed, but to simplify the design, the TOC bar will not be shown even when this button scrolls out of the top.

    https://github.com/user-attachments/assets/0095fdd1-2cb7-489f-b24a-d6c28ba0635c

cotes2020:
> why there is such a difference between mobile and tablet mode? 

Reduce your workload. If you think the tablet maintains the same UX as the phone, I support that too.

kungfux:
FYI: The design solution is ready for preview.

cotes2020:
In terms of UI design, we also need to pay attention to maintaining the responsive variation of the padding on the left and right sides of the secondary top bar. Currently, this aspect is lacking.

https://github.com/user-attachments/assets/72960706-ebaf-481f-a6d7-b2668f8269bc

Note: In the video above, I used the browser's **Developer Tools** to keep the secondary top bar always visible to illustrate the situation.

> What do you think if we place breadcrumb on the secondary panel? ...

I don't think it's necessary. First, the horizontal space in the top bar on mobile devices is very narrow, so we should leave as much space as possible for the title. Second, in mobile browsers, you can simply swipe left to return to the previous page.

cotes2020 (commented):
I've made some changes to this PR and will submit a commit to the same branch for your review shortly.

kungfux:
@cotes2020 Thanks for taking care of the styles and TOC generation. Is there anything else you would like to change?

cotes2020:
Yes, I am working on fixing the issue with the TOC bar flickering when toggling between show/hide in iOS Safari.

cotes2020 (approved):
I've completed the improvements, do you want to test them? If there are no issues, it's ready to be merged.

cotes2020 (commented):
Wait, maybe it's better to leave it to test for a week or so.

kungfux:
I like the result you achieved. A few things to note though:
- Init TOC once only leads to the problem for desktop users that if the browser window is resized or pinned, then the "Contents" button will not trigger the popup / or the TOC will not appear in the sidebar. This could potentially happen on mobile devices when the orientation changes.
- Using different icons for buttons that do exactly the same thing may not be the best UX choice.
- The scroll bar does not look good on the popup. It may be better not to make the header scrollable, just the TOC.

cotes2020:
> Init TOC once only leads to the problem for desktop users that if the browser window is resized or pinned, ...

The tocbot plugin has an issue where creating multiple instances causes scrolling tracking to fail. Currently, this bug is known to trigger only on tablets with a screen width > 1200px. The manual solution is to refresh the page to run `tocbot.init()` again. Requiring users to refresh the page feels a bit silly, and I’m quite troubled by this. So, do you have a better solution?

> Using different icons for buttons that do exactly the same thing may not be the best UX choice.

To clarify, the separate TOC solo button is not designed to mimic the appearance of the TOC top bar. The label "Contents" serves the same purpose as the icon on the right side of the TOC bar, while the ">" icon is just a symbol that implies "expansion," indicating that clicking this button will reveal something on the right. I insist on the current design as it is more concise and does not take up too much attention :)

> The scroll bar does not look good on the popup. It may be better not to make the header scrollable, just the TOC.

Thank you for bringing up this important issue! I hadn't noticed that forcefully scrolling down on the popup window in mobile browsers also moves the header. I have now corrected it.

kungfux:
> The tocbot plugin has an issue where creating multiple instances causes scrolling tracking to fail. Currently, this bug is known to trigger only on tablets with a screen width > 1200px. The manual solution is to refresh the page to run `tocbot.init()` again. Requiring users to refresh the page feels a bit silly, and I’m quite troubled by this. So, do you have a better solution?

If I understand the problem correctly, the possible solution might be to track window resize and call `tocbot.refresh()`. I see no tracking problems with the example below.

<details>
 <summary>Code example</summary>

```js
 // Make sure to remove `if (window.innerWidth < 1200)` to execute all init functions in `post.js`
 // Handle TOC layout switch from desktop to mobile and vise versa
  const mediaQuery = matchMedia('(max-width: 1200px)');
  function refreshToc(e) {
    if (e.matches) {
      console.log('< 1200');
      tocbot.refresh({
        tocSelector: '#toc-popup-content',
        contentSelector: '.content',
        ignoreSelector: '[data-toc-skip]',
        headingSelector: 'h2, h3, h4',
        orderedList: false,
        scrollSmooth: false,
        collapseDepth: 4,
        headingsOffset: barHeight
      });
    } else {
      console.log('> 1200');
      tocbot.refresh({
        tocSelector: '#toc',
        contentSelector: '.content',
        ignoreSelector: '[data-toc-skip]',
        headingSelector: 'h2, h3, h4',
        orderedList: false,
        scrollSmooth: false,
        headingsOffset: 16 * 2 // 2rem
      });
    }
  }
  mediaQuery.addEventListener('change', refreshToc);
```
</details>

_javascript/modules/components/toc.js
```diff
@@ -1,15 +1,29 @@
...
-
-    document.getElementById('toc-wrapper').classList.remove('d-none');
+import { TocMobile as mobile } from './toc/toc-mobile';
+import { TocDesktop as desktop } from './toc/toc-desktop';
+
+const mediaQuery = matchMedia('(min-width: 1200px)');
+
+function refresh() {
+  if (mediaQuery.matches) {
+    desktop.refresh();
```
kungfux, line 8: Please call `mobile.hide();` to avoid situation when viewport gets resized and we get empty popup window. I guess it will be fine to just close it since TOC will appear on the right sidepanel.

```suggestion
    mobile.hide();
    desktop.refresh();
```

cotes2020: It makes sense. :)

_javascript/modules/components/toc/toc-desktop.js
```diff
@@ -0,0 +1,22 @@
+export class TocDesktop {
+  /* Tocbot options Ref: https://github.com/tscanlin/tocbot#usage */
+  static options = {
```
kungfux, line 3: "How to change TOC configuration?" is quite popular community question. The [solution](https://github.com/cotes2020/jekyll-theme-chirpy/discussions/1364#discussioncomment-10812644) I proposed will become more complex after this feature. I wonder if there is anything we can do to make it easier? Maybe you have an idea?

_sass/layout/post.scss
```diff
@@ -228,6 +228,7 @@ header {
   }
 }
 
+/* TOC pannel */
```
kungfux, line 231:
```suggestion
/* TOC panel */
```

_javascript/modules/components/toc/toc-mobile.js
```diff
@@ -0,0 +1,85 @@
...
+      this.initComponents();
+    }
+
+    tocbot.refresh(this.options);
+  }
+
+  static showPopup() {
+    TocMobile.setScrollEnabled(false);
+    popup.showModal();
+  }
```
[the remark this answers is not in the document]
cotes2020: It's better to center it.

_javascript/modules/components/toc/toc-desktop.js
cotes2020: There are ways to do this, such as reading tocbot's configuration from an external file, but this will run down performance and make the project bigger and bigger — [Feature Creep](https://en.wikipedia.org/wiki/Feature_creep).

I would suggest not to worry too much about this kind of customization, if they have a personal need, the easiest way is to change the configuration in the theme source code, and then compile a `post.min.js` and copy it to their Starter project.

kungfux: I was thinking more in the direction of extracting the options definition into a separate object that can be easily accessed and modified from within a custom script, for example. That way it will have no runtime impact while still being customizable.

cotes2020: No matter how much we tweak it, we still have to maintain a piece of ugly "customization" code, don't we? I think the main goal of the project is to maintain a basic design, and since all the source code is open-source, "customization" should be the responsibility of users who are dissatisfied with parts of the original design.

Instead of exposing the config options for tobot, I think a more feasible approach would be to add a "expand/collapse" button for the TOC on the desktop view. But,  it is a topic outside the scope of this PR.

cotes2020 (approved):
As a note, adding `.overflow-hidden` to the `body` element will not disable scrolling on iOS Safari, but I don't think that's important.

PR #1990, opened 2024-10-12
fix: close popup gracefully with Esc key

kungfux:
## Type of change

## Description
**Problem:** Closing a popup by pressing `Esc` is not handled correctly.

**Scenario:**
- Open blog post on mobile view
- Click `Contents`
- Close popup with `Esc` key

**Expected:** Popup is closed with animation.
**Actual/Problem:** Popup is closed without animation.

- Click `Contents` again

**Expected:** Popup can be opened again.
**Actual/Problem:** Popup is being closed on open (for the first time after closing). See recording attached.

<details>
<summary>Recording</summary>

popup-esc
</details>

Ref #1964

### DIFF
--- a/_javascript/modules/components/toc/toc-mobile.js
+++ b/_javascript/modules/components/toc/toc-mobile.js
@@ -62,7 +62,11 @@ export class TocMobile {
     activeItem.scrollIntoView({ block: 'center' });
   }
 
-  static hidePopup() {
+  static hidePopup(event) {
+    if (event?.type === 'cancel') {
+      event.preventDefault();
+    }
+
     if (!$popup.open) {
       return;
     }


_javascript/modules/components/toc/toc-mobile.js
```diff
@@ -62,7 +62,9 @@ export class TocMobile {
     activeItem.scrollIntoView({ block: 'center' });
   }
 
-  static hidePopup() {
+  static hidePopup(event) {
+    event?.preventDefault();
+
```
cotes2020, line 67: This doesn't make it clear what the intent of this line of code is, how about adding a specific condition `event.type === 'cancel'`?

```suggestion
  static hidePopup(event) {
    if (!$popup.open) {
      return;
    }

    if (event && event.type === 'cancel') {
      event.preventDefault();
    }
```

cotes2020 (changes requested):

cotes2020 (approved):

PR #2053, opened 2024-11-21
Allow images to be hidden from the top of posts

mattpopovich:
## Type of change

- [x] Documentation update

## Description

I was looking for a way to hide a post's image from the top of the post but still have it show on the home page and whenever it is shared in socials. This was mainly because I wanted the first part of the post to embed a corresponding YouTube video and I didn't want the post image and YouTube video to be displayed right next to each other, effectively showing the same things twice. 

I found [this discussion post](https://github.com/cotes2020/jekyll-theme-chirpy/discussions/1804) from someone else that is asking how to do this exact thing, and followed the suggestions there and made it into a PR.

## Additional context

This PR successfully hides the image from being shown at the top of the post.

- [ ] However, I'm unsure if this prevents the image from being shown when sharing via socials. It seems to be working via a "social share preview" browser extension but I'm not sure where `og:image` is being set, so please confirm that `og:image` is set correctly before merging.

### DIFF
--- a/_layouts/post.html
+++ b/_layouts/post.html
@@ -45,12 +45,15 @@ script_includes:
           {%- capture lqip -%}lqip="{{ page.image.lqip }}"{%- endcapture -%}
         {% endif %}
 
-        <div class="mt-3 mb-3">
-          <img {{ src }} {{ class }} {{ alt }} w="1200" h="630" {{ lqip }}>
-          {%- if page.image.alt -%}
-            <figcaption class="text-center pt-2 pb-2">{{ page.image.alt }}</figcaption>
-          {%- endif -%}
-        </div>
+        {% if page.image.show_image_in_post != false %}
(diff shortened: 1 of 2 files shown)


cotes2020 (commented):
This change complicates the preview image and may cause confusion. Instead of hiding the top image in the post body, it would be better to display the embedded video as a link.

For the `og:image` issue, I don't have the energy to delve into it right now, so you'll have to resolve it yourself. You can check the code in `_includes/head.html` and the API of the `jekyll-seo-tag` plugin, and I'm sure you'll find some clues.

mattpopovich:
Ok. I hoped that by making it an "opt-in" change (if you don't want to use it, no changes are needed), that would reduce confusion. 

Sounds like you are not interested in integrating this to the baseline, so I will close this PR.

PR #2071, opened 2024-11-29
fix: ensure pageviews are fetched after DOM is loaded

kungfux:
## Type of change

## Description
The pageviews script is loaded inside the `<head>` and is therefore executed too early, causing pageviews to display an infinite spinner. In Chirpy 7.1.1 the script was loaded at the end of the `<body>`.

image


cotes2020 (approved):
Nice fix, the pageview part was missing when I changed the JS at `v7.2.0`

PR #2072, opened 2024-11-30
chore(ui): toggle sidebar icon on sidebar open

kungfux:
## Type of change

## Description
Change the sidebar icon to 'close' when the sidebar is open to make the purpose of the button clear.
This is a nice to have improvement, as suggested in #2059.

image
image


cotes2020 (commented):
Hi, @kungfux,

Please don't modify the original design, the switch icon doesn't need to change dynamically.

It is true that some designs favor the use of an 'x', but then the 'x' design logically belongs to the sidebar, not in the original switch position.

PR #2134, opened 2024-12-18
Remove the `aspect-ratio` directive from `.embed-video` CSS class

seanthegeek:
## Type of change

## Description

Not all video is in 16:9 aspect ratio (e.g., vertical video), and forcing that causes the videos to appear small with large black spaces on either side.

### DIFF
--- a/_sass/base/_base.scss
+++ b/_sass/base/_base.scss
@@ -291,7 +291,6 @@ main {
   width: 100%;
   height: 100%;
   margin-bottom: 1rem;
-  aspect-ratio: 16 / 9;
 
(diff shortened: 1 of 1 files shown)


cotes2020 (commented):
This is a good point. When designing, I envisioned embedded videos to be in landscape mode, so a 16:9 aspect ratio is sufficient. In the future, a new parameter might be introduced to accept custom classes to accommodate the aspect ratio of short videos.

However, simply removing the original 16:9 aspect ratio in your modification is not advisable, as it will break the existing functionality.

PR #2186, opened 2025-01-12
Update gitignore and ignore HTTP links

godylockz:
## Type of change
- [X] Improvement (refactoring and improving code)
- [ ] Documentation update

## Description
Git Ignore Additions - Add cache/obsidian
Ignore HTTP links during building

### DIFF
--- a/.github/workflows/starter/pages-deploy.yml
+++ b/.github/workflows/starter/pages-deploy.yml
@@ -53,6 +53,7 @@ jobs:
       - name: Test site
         run: |
           bundle exec htmlproofer _site \
+            \-\-no-enforce-https \
             \-\-disable-external \
             \-\-ignore-urls "/^http:\/\/127.0.0.1/,/^http:\/\/0.0.0.0/,/^http:\/\/localhost/"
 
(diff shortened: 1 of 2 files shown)


cotes2020 (commented):
`.gitignore` has different configurations in different environments. Currently, this project does not have the files you listed.

If `html-proofer` does not enforce HTTPS, it will compromise the security of network transmission because the output page contains HTTP links.

godylockz:
Figured I'd suggest these changes just in case I'm not the only one in the community with this type of configuration.

> `.gitignore` has different configurations in different environments. Currently, this project does not have the files you listed.

Obsidian is a popular markdown editor. If I open the folder as a vault, would want to ignore it from commit history as it changes every time the vault is opened.

> If `html-proofer` does not enforce HTTPS, it will compromise the security of network transmission because the output page contains HTTP links.

Modern browsers default HTTPS when available. If I'm making a blog post with a http link, for instance if the target URL is only available on port 80 or another port then the proofer needs to be setup for ignoring or it won't build.

cotes2020:
I understand your motivation, but ultimately most of the modifications are specific to your use case. Although I do not use Obsidian , it is acceptable to add its hidden folder to `.gitignore`. However, please keep the other modifications in your branch.

PR #2243, opened 2025-02-09
feat(direction): add direction option

davidazad1:
## Type of change

- [ ] Documentation update

## Description

This pull request introduces support for right-to-left (RTL) language direction in the Jekyll Chirpy theme. The changes include:

1. Added a new direction configuration option in _config.yml to allow users to set the direction of the webpage (ltr for left-to-right or rtl for right-to-left).
2. Updated multiple template files (_includes/*.html and _layouts/default.html) to dynamically apply the dir attribute based on the site.direction setting. This ensures proper rendering for RTL languages.

The default direction is set to ltr if the direction option is not specified in _config.yml.

## Additional context

This addiction does not introduce any breaking changes and is fully compatible with existing functionality. The direction can set by setting the direction parameter in the site configuration.


cotes2020 (changes requested):
Thank you for implementing this feature. However, there still seems to be quite a bit of work to do to complete it:

1. The elements in the sidebar do not have consistent right margins, and the spacing of the social icons at the bottom is uneven.
2. The positions of the breadcrumb and search input box in the top bar are reversed.
3. The metadata in the homepage article list does not follow the right-to-left order.



3. The dates on the archive page are stacked together, and the vertical line of the timeline is also separated from the entries on the left and right.

davidazad1:
Hi @cotes2020, I've fixed the issues you pointed out. Please take a look when you have time. Thanks!

davidazad1:
Hi @cotes2020, I've fixed the issues you pointed out. Please take a look when you have time. Thanks!

cotes2020:
If you don't mind, please upload the modified rendering result image so that we can review it more quickly. Thank you.

davidazad1:
ok
image
image

cotes2020 (changes requested):
Although it's an improvement over the first version, the recent updated and trending tags on the right side are still noticeably to the right, as you may have noticed.

Also, you've added a lot of invalid commits to this PR, please remove them, thanks.

PR #2330, opened 2025-04-01
Update search-results.html

alazaroc:
Fix value of "trending_tags" text in multi-language in mobile

## Type of change

## Description
The text for "trending_tags" was displaying incorrectly on mobile devices when using multi-language support.

### DIFF
--- a/_includes/search-results.html
+++ b/_includes/search-results.html
@@ -3,7 +3,7 @@
 <div id="search-result-wrapper" class="d-flex justify-content-center d-none">
   <div class="col-11 content">
     <div id="search-hints">
-      {% include_cached trending-tags.html %}
+      {% include_cached trending-tags.html lang=lang %}
     </div>
     <div id="search-results" class="d-flex flex-wrap justify-content-center text-muted mt-3"></div>
   </div>
--- a/_includes/search-results.html
+++ b/_includes/search-results.html
@@ -3,7 +3,7 @@
 <div id="search-result-wrapper" class="d-flex justify-content-center d-none">
   <div class="col-11 content">
     <div id="search-hints">
-      {% include_cached trending-tags.html lang=lang %}
+      {% include_cached trending-tags.html lang=include.lang %}
     </div>
     <div id="search-results" class="d-flex flex-wrap justify-content-center text-muted mt-3"></div>
   </div>


RlxChap2 (commented):
perfect 👌👌

cotes2020:
Does it mean that after modifying the `lang` in `_config.yml`, the **Trending Tags** did not change? If so, it seems that this issue does not exist.

cotes2020 (commented):
Does it mean that after modifying the `lang` in `_config.yml`, the **Trending Tags** did not change? If so, it seems that this issue does not exist.

alazaroc:
Exactly @cotes2020. If I configure Spanish as my language in `_config.yml` and try to access the search on the mobile version (only on the mobile version), the English text is shown:
image

If you include the line that I put in this PR, this is fixed. This other image shows the fix:
image

Furthermore, all other HTML page inclusions utilize `lang=lang`, except for this particular file

cotes2020:
Sorry, I still can't reproduce the error you mentioned.  

`include_cached` has already cached `trending-tags.html` in `_layouts/default.html`, so it's fine even if `search-results.html` doesn't specify `lang`. 

Of course, for better consistency, adding the `lang` parameter in `search-results.html` wouldn't hurt.

alazaroc:
mmm, then this might be due to other modifications in my own version. I also have multi-language enabled (en/es). Never mind, I've already made the change on my version, it's your call here

_includes/search-results.html
```diff
@@ -3,7 +3,7 @@
 <div id="search-result-wrapper" class="d-flex justify-content-center d-none">
   <div class="col-11 content">
     <div id="search-hints">
-      {% include_cached trending-tags.html %}
+      {% include_cached trending-tags.html lang=lang %}
```
cotes2020, line 6: > ...then this might be due to other modifications in my own version.

Indeed, the issue does not exist in the `production` branch, you can also give it a try.

Additionally, if you want to add parameters here, you should pass the parameters from the upper level, not just `lang`.

```suggestion
      {% include_cached trending-tags.html lang=include.lang %}
```

cotes2020 (changes requested):

cotes2020 (approved):

PR #2368, opened 2025-04-21
feat(contact): Adds GitLab for sidebar. Only display contacts that are configured.

JasonRaveling:
## Type of change

## Description
Adds GitLab as an option, similar to GitHub. It is an optional config. When set, it will show the GitLab icon in the sidebar with a link to the GitLab page of the configured username.


JasonRaveling:
This started as just a simple add of GitLab. Then I realized certain contacts (e.g. Twitter) were being added even when none was configured. This pull request also includes an adjustment to the conditional so that if a username is not set for GitHub or Twitter (also GitLab), then it will not be displayed in the sidebar.

cotes2020 (changes requested):
Actually, we don't need to add any configuration in `_config.yml`; we just need to add gitlab like in the example below.

https://github.com/cotes2020/jekyll-theme-chirpy/blob/401e2af0f8a173d8437e03027c7aff558e8c0bde/_data/contact.yml#L34-L36

JasonRaveling:
Thanks for pointing that out. After looking a little closer, `username` is only used to generate the URLs for Github and Twitter. Would it make sense to remove those from `_config.yml`? Or are those still there for backward compatibility or something else?

cotes2020:
Maybe you didn't notice the comments.

https://github.com/cotes2020/jekyll-theme-chirpy/blob/d0f8f9553e41536eb84ae2fdd3f3bc9d13f7ef8c/_config.yml#L14-L15

JasonRaveling:
Closing out since it will not be implemented. Looking in to adding the adjustment to the conditional so that if a username is not set for GitHub or Twitter (also GitLab), then it will not be displayed in the sidebar.

PR #2407, opened 2025-05-17
Fix image shadow not applied inside posts and add theme-based shadow color support

Minssuy99:
## Type of change

## Description
Fixed a bug where the `.shadow` class was not applied to images inside post content.

Defined the missing `.shadow` class and added theme-based `--img-shadow` variables in `_light.scss` and `_dark.scss` to allow different shadow colors depending on the active theme.

## Additional context
None

shadow bug fix

### DIFF
--- a/_sass/base/_base.scss
+++ b/_sass/base/_base.scss
@@ -474,3 +474,10 @@ mjx-container {
     margin-top: 3rem;
   }
 }
+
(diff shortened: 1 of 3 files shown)


cotes2020:
This modification makes the shadow color even worse, let it stay in your fork.

PR #2420, opened 2025-05-27
fix: correct pageviews extraction

kungfux:
## Type of change

## Description
**Problem:** `NaN` appears for pageview when over 1K pageview's value is returned by GoatCounter as JSON value contains formatted value like `1,234` that is expected according to [documentation](https://www.goatcounter.com/help/visitor-counter#json-160).

**Solution:** Remove any non-digit symbols from the value before trying to parse it.

## Additional context
Reported by @SoumyaK4, discussed and tested under #2412.

### DIFF
--- a/_includes/pageviews/goatcounter.html
+++ b/_includes/pageviews/goatcounter.html
@@ -10,7 +10,7 @@
       fetch(url)
         .then((response) => response.json())
         .then((data) => {
-          const count = data.count.replace(/\s/g, '');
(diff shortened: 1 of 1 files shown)


cotes2020 (approved):
Thanks for the fix. GoatCounter seems to use a space or comma as the thousands separator.

```console
$ curl https://chirpy-demo.goatcounter.com/counter/%2Fposts%2Fgetting-started.json
{
  "count": "34 640",
  "count_unique": "34 640"
}
```

PR #2477, opened 2025-07-21
improving giscus repo comment in _config.yml

minxsecurity:
## Type of change

- [x] Documentation update

## Description
updated comment next to giscus repo parameter to help others know that it should be the same repo as the github pages repo. otherwise you will get content security policy errors and it won't load properly.

I mistakenly was under the impression that you had to fork/clone the giscus repo into your personal repository, and use that instead, and wanted to hopefully help out someone else in case they had the same thought.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -119,7 +119,7 @@ comments:
     issue_term: # < url | pathname | title | ...>
   # Giscus options › https://giscus.app
   giscus:
-    repo: # <gh-username>/<repo>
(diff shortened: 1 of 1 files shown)


cotes2020:
Thank you for your PR! However, I don't think the original comment needs to be changed. In fact, you can create a separate repository for storing Giscus comments for your site. If you have more questions, please refer to the documentation on <https://giscus.app/>.

PR #2506, opened 2025-08-25
refactor(sidebar contact): Adds a conditional to check if a GitHub or…

JasonRaveling:
… Twitter username has been configured before displaying the icon in as a contact.

Ensures a GitHub or Twitter username is configured in `_config.yml` before displaying them in the sidebar. Adds a conditional to capture/assign `url` so that it is `nil` if no username was provided. That way when the check for `url` is done for building the link, it will be skipped if no username is configured.

## Type of change

## Description
Before this commit, GitHub and Twitter icons would be displayed on the sidebar, even if you deleted the username from `_config.yml`. In that case, the logo would just appear with a link to the root site (e.g. twitter.com instead of twitter.com/some_user). This fix adds logic so that when `_includes/sidebar.html` is rendered it will check if `_config.yml` has an entry for it. If no entry is configured the icon will not be displayed.


JasonRaveling:
Closing because I suspect it will never pass commit linting. Rebasing to amend the commit message doesn't seem to haven helped.

PR #2584, opened 2025-11-09
feat: add pagination to search results

### LINKED ISSUE #2583: Make search result limit configurable
### Checklist

### Is your feature request related to a problem? Please describe

The built-in search only shows the first 10 matches. When a query has more matches, there is no pagination or way to reveal the remaining posts, so visitors cannot discover older content.

### Describe the solution you'd like

Expose a configuration (for example via `_config.yml` or a dedicated include) so that the `limit` option passed to `SimpleJekyllSearch` can be set by site owners, or implement pagination/infinite scroll so users can browse beyond the first page of results.

### Describe alternatives you've considered

- Override `_includes/search-loader.html` locally to pass a higher `limit` value. This works but duplicates theme code and must be maintained by each site.
- Replace the search with an external service such as Algolia or Elasticlunr, which increases setup complexity.

### Additional context

- Chirpy version: 7.3.1 (`bundle show jekyll-theme-chirpy`)
- The theme initializes `SimpleJekyllSearch` without overriding its default `limit` (10) in `_includes/search-loader.html`.
- `Simple-Jekyll-Search` documents the default limit of 10 results: https://github.com/christian-fei/Simple-Jekyll-Search/blob/master/dest/simple-jekyll-search.js#L311-L353
- Steps to reproduce:
  1. Build a site with more than 10 posts that match the same query.
  2. Open the search modal and type the query.
  3. Only the first 10 posts are rendered, and there is no option to view additional matches.

KKamJi98:
## Type of change

## Description
- mirror the home page paginator UI in the search panel so users can browse search results page by page
- cache raw SimpleJekyllSearch output and render only the active page, keeping the DOM small and the behavior consistent on desktop/mobile (including ellipsis + prev/next arrows)
- expose `search.limit` and `search.per_page` in `_config.yml` so site owners can control how many hits are fetched and shown per page
- added regression coverage by running `bash tools/test.sh`

## Additional context
- Fixes #2583


KKamJi98:
Rebased onto latest `master` and squashed into a single commit — all checks are green. Whenever you have a moment, I'd appreciate a review. Happy to adjust anything. Thanks!

PR #2614, opened 2025-12-03
feat(giscus): allow giscus data-loading to be eager (not lazy)

joshlk:
## Type of change
- [X] New feature (non-breaking change which adds functionality)

## Description
Add a config option for Giscus comments, which allows you to set the iframe data-loading. Currently, it's hard-coded for lazy loading, but this will enable you to set it to load eagerly when the page loads. I personally don't think it's a good user experience for the user to wait for the comments to load once they scroll to the bottom of the page. The default has been kept as lazy loading.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -129,6 +129,7 @@ comments:
     input_position: # optional, default to 'bottom'
     lang: # optional, default to the value of `site.lang`
     reactions_enabled: # optional, default to the value of `1`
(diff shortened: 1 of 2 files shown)


cotes2020 (changes requested):
`data-loading='eager'` is the default behavior, so there's no need to set it manually. Using a `site.comments.giscus.lazy_loading: true` option would be more intuitive and concise.

PR #2642, opened 2026-01-13
fix: sidebar email link

lynkos:
## Type of change

## Description
Clicking the **Mail** icon in Firefox redirects the user to a new blank page that displays the `mailto` address. See the **Additional context** section for screenshots. Note: This bug doesn't appear in Chrome or Safari.

This change doesn't require any additional dependencies since it's a one-line change in `_includes/sidebar.html`.

It also doesn't break the behavior in Chrome or Safari (i.e. no new bug(s) introduced) since this bug only affects Firefox, to my knowledge.

Please refer to [my fork](https://blog.lynkos.dev) and click the **Mail** icon for a working example of my fix (i.e. prevent site from redirecting to a blank page after clicking the **Mail** icon).

## Additional context
Clicking the **Mail** icon in Firefox
Screenshot of bug

Firefox user is redirected to this page after clicking the icon
Screenshot of bug


cotes2020 (changes requested):
Firefox handles `javascript:` URLs such that if the expression does not return `undefined`, it treats the return value as page content and displays it, which causes navigation to a blank page showing the `mailto` address. Chrome and Safari do not handle it this way, so they are not affected.

The fix is to wrap the expression in `void()`, ensuring it returns undefined:

```js
javascript:void(location.href = 'mailto:' + ['{{ email[0] }}','{{ email[1] }}'].join('@'))
```

cotes2020 (changes requested):

cotes2020 (approved):

PR #2648, opened 2026-01-23
Fix _include/media-url.html logic

### LINKED ISSUE #2639: Site generation duplicates baseurl for img when baseurl not empty
### Checklist

### Describe the bug

The problem manifest itself when `baseurl` in the `_config.yml` is defined as something other than empty.
This problem affects all markdown where an `` tag is generated. Whatever the `src` attributed contains, it gets the `baseurl` prepended twice.

This seems to happen because a final step in rendering is to pull things through the refactoring code. For `` tags, this then results in the subsequent code in refactoring ultimately calling
```
{% include media-url src=<baseurl>/assets/images/Flag.png> %}
```
This shows that the baseurl was already applied before it got to this point. The include prepends it again.

### Steps To Reproduce

1. Define baseurl as "my_site" in _config.yml
2. In any `.md` file include (thec second variation to show this problem has nothing to do with the jekyll_picture_tag plugin):
```
Picture plugin: {% picture "Flag.png" --alt Flag %}

MD Image full path: ![Test]({% link /assets/images/Flag.png %})
```
3. Observe output (formatted for clarity):
```
<p>
	Picture plugin:
	<picture>
		<source srcset="/my_site/assets/images_gen/Flag-400-dd43f229f.webp 400w, /my_site/assets/images_gen/Flag-600-dd43f229f.webp 600w, /my_site/assets/images_gen/Flag-800-dd43f229f.webp 800w, /my_site/assets/images_gen/Flag-1000-dd43f229f.webp 1000w, /my_site/assets/images_gen/Flag-1600-dd43f229f.webp 1600w" type="image/webp" />
		<source srcset="/my_site/assets/images_gen/Flag-400-a5c1e79ca.png 400w, /my_site/assets/images_gen/Flag-600-a5c1e79ca.png 600w, /my_site/assets/images_gen/Flag-800-a5c1e79ca.png 800w, /my_site/assets/images_gen/Flag-1000-a5c1e79ca.png 1000w, /my_site/assets/images_gen/Flag-1600-a5c1e79ca.png 1600w" type="image/png" />
		<a href="/my_site/my_site/assets/images_gen/Flag-800-a5c1e79ca.png" class="popup img-link shimmer">
			<img src="/my_site/my_site/assets/images_gen/Flag-800-a5c1e79ca.png" alt="Flag" loading="lazy">
		</a>
		</picture> 
</p>
<p>
	MD Image full path:
	<a href="/my_site/my_site/assets/images/Flag.png" class="popup img-link shimmer">
		<img src="/my_site/my_site/assets/images/Flag.png" alt="Test" loading="lazy">
	</a> 
</p>
```
4. Notice that in both cases the `` tag was surrounded by an `<a>` during refactoring (fine), **and** the `src` attribute in the tag now has the `baseurl` value in there twice. Tags like `<source>` which also contain image paths are unaffected by the refactoring code, and consequently are correct.

### Expected Behavior

Do not double up on the baseurl

### Environment

- Ruby: 
- Jekyll: 
- Chirpy: 

### Anything else?

_No response_

dolfs:
## Type of change

## Description
When `baseurl` in `_config.yml` is not empty, without this fix, it can be prepended more than once, which is incorrect.

This fix adds logic to prevent prepending the `baseurl` when it is already a prefix in the url, but correctly handles cases where it is a prefix, but the url does not follow this prefix with a path/url separator.

For example, when `baseurl` is set to `/blog`, but the url starts with `/blogger` it is not considered "prefixed" and `/blog/ is prepended (resulting in `/blog/blogger…`, but if the url starts with `/blog/` it is considered "prefixed" and is not prepended again (keeping/resulting in `/blog/…`)

A limitation is that if the user supplied a (partial) URL starting with `/blog/` and truly intended to produce `/blog/blog/`, it won't work.

## Additional context
Fixes #2639, and:

* Prevents occasional introduction of white space
* Check for pre-existing absolute url is made more robust
* Better handling of scenarios where site.baseurl ends in '/', or accidentally was set to just '/'

### DIFF
--- a/_includes/media-url.html
+++ b/_includes/media-url.html
@@ -2,36 +2,86 @@
   Generate media resource final URL based on `site.cdn`, `page.media_subpath`
 
   Arguments:
-    src - required, basic media resources path
-    subpath - optional, relative path of media resources
-    absolute - optional, boolean, if true, generate absolute URL
+    src       - required, basic media resource path
+    subpath   - optional, relative path of media resources
+    absolute  - optional, boolean, force full absolute URL
 
   Return:
     media resources URL
 {%- endcomment -%}
 
-{% assign url = include.src %}
+{%- assign url = include.src -%}
 
-{%- if url -%}
-  {% unless url contains ':' %}
-    {%- comment -%} Add media resources subpath prefix {%- endcomment -%}
-    {% assign url = include.subpath | default: '' | append: '/' | append: url %}
-
-    {%- comment -%} Prepend CND URL {%- endcomment -%}
-    {% if site.cdn %}
-      {% assign url = site.cdn | append: '/' | append: url %}
-    {% endif %}
-
-    {% assign url = url | replace: '///', '/' | replace: '//', '/' | replace: ':/', '://' %}
-
-    {% unless url contains '://' %}
-      {% if include.absolute %}
-        {% assign url = site.url | append: site.baseurl | append: url %}
-      {% else %}
-        {% assign url = site.baseurl | append: url %}
-      {% endif %}
-    {% endunless %}
-  {% endunless %}
+{%- comment -%}
+  Normalize site.baseurl:
+  - strip whitespace
+  - treat "/" as empty (common misconfiguration)
+  - remove trailing slash for other non-empty values
+{%- endcomment -%}
+{%- assign b_url = site.baseurl | default: '' | strip -%}
+{%- if b_url == '/' -%}
+  {%- assign b_url = '' -%}
+{%- elsif (b_url | slice: -1, 1) == '/' -%}
+  {%- assign b_url = b_url | remove_last: '/' -%}
 {%- endif -%}
+
+{%- comment -%}
+  Normalize CDN:
+  - strip whitespace
+  - remove trailing slash
+{%- endcomment -%}
+{%- assign cdn = site.cdn | default: '' | strip -%}
+{%- if cdn != '' and (cdn | slice: -1, 1) == '/' -%}
+  {%- assign cdn = cdn | remove_last: '/' -%}
+{%- endif -%}
+
+{%- comment -%}
+  Only process non-absolute URLs
+{%- endcomment -%}
+{%- unless url contains '://' -%}
+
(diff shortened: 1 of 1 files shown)


cotes2020 (changes requested):
Hi there, thanks for the PR and for your time!

From the first review, I’ve identified two issues:

1. **`remove_last`** is a feature of **Liquid 5.2.0**, but Jekyll currently only supports **Liquid 4.x**.
2. There is a **bug** in this PR: when `site.cdn` is not empty, the output media URL results in `/<baseurl><cdn>/<subpath>`.

dolfs:
Changes made.

cotes2020:
> Changes made.

Where is it? Did you forget to push?

dolfs:
Oops, it should be there no. I am not very experienced with PRs.

cotes2020:
After looking deeper into the related issue #2639, I don't think this change/PR is necessary. `media-url.html` is a snippet that is explicitly called, and whether the `src` parameter includes the `baseurl` is something the caller can control. Adding logic to detect the `baseurl` inside the snippet would not only overcomplicate it but, as you mentioned, we can't be certain if the `src` path just happens to match the `baseurl`. 

> A limitation is that if the user supplied a (partial) URL starting with /blog/ and truly intended to produce /blog/blog/, it won't work.

cotes2020:
I've come up with a workaround: when using the `link` tag or other third-party filters, we can append a specific class (e.g., `.remain`) to the image. This acts as a flag to tell the `media-url` filter to skip any modifications.

Change `media-url.html` to:
```liquid
{%- assign remain = include.remain | default: false -%}

{%- if url -%}
  {%- unless url contains '://' or remain -%}
    {%- comment -%}
      Handle src URL using existing logic.
      ...
    {%- endcomment -%}
  {%- endunless -%}

{{- url -}}
```

Usage:
```markdown
![alt](/path/to/img)  <!-- output: <img src='[site.cdn]/[site.baseurl]/path/to/img'> -->

![alt]({% link /path/to/img %}){: .remain } <!-- output: <img src='/[site.baseurl]/path/to/img'> -->
```

PR #2717, opened 2026-04-13
Update the Google Analytics file to match the recommended snippet by Google

marcrevelles:
## Type of change

## Description
I was setting up Google Analytics on my personal blog, which uses the Chirpy theme, and I realized that the content in `_includes/analytics/google.html` seems to be outdated.

The Google Analytics tag setup page recommends the following code, which is present in the pull-request. I have put the template variable in the right place and tested the new setup.

Proof:
> ga_tag_setup

## Additional context
I have read the Contributing Guideline. However, this is my first real contribution to an open source project, so I may have missed something. Tell me if it’s necessary that I create a feature request issue.


cotes2020 (commented):
The original code is identical to Google's, except that it uses lazy loading.

PR #2764, opened 2026-07-01
feat(plugins): add Knocket live chat widget

rrivory:
## Type of change

## Description

Add Knocket as an optional live chat service, leveraging the existing analytics provider auto-include pattern in `head.html`.

Knocket is a 100% free live chat widget — free forever, no ads, no seat limits. Disabled by default; renders only when `analytics.knocket.id` is set in `_config.yml`.

Config example:

```yaml
analytics:
  knocket:
    id: YOUR_ID   # from https://trtc.io/solutions/knocket
```

More info: https://trtc.io/solutions/knocket

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -73,6 +73,8 @@ analytics:
     id: # fill in your Cloudflare Web Analytics token
   fathom:
     id: # fill in your Fathom Site ID
+  knocket:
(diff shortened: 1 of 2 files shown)


cotes2020:
Thanks, but this project won't be adopting this plugin.

PR #545, opened 2022-04-02
feat: horizontal scroll for long equations

Jayitha:
# Description

Currently, long equations rendered by MathJax tend to overflow the post content onto the right side navigational toc and on phones, they get completely cut off. PR introduces a horizontal scroll for along equations.  

Fixes #543 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## How has this been tested

I specified a long equation detailed below

```markdown
$$
(a + b)^3 = {3 \choose 0} a^3 b^0 + {3 \choose 1} a^2 b^1 + {3 \choose 2} a^1 b^2 + {3 \choose 0} a^0 b^3 + 0 \left ({3 \choose 0} a^3 b^0 + {3 \choose 1} a^2 b^1 + {3 \choose 2} a^1 b^2 + {3 \choose 0} a^0 b^3 \right )
$$
```

This equation currently runs over the navigation toc in the browser and gets completely cut off on the phone as described in the feature request: #543 

Post making the changes in this PR, a disappearing horizontal scroll appears for long equation. 

- [x] I have run `bash ./tools/deploy.sh --dry-run` (at the root of the project) locally and passed
- [x] I have tested this feature in the browser

### Test Configuration

- Browser type & version: Safari (15.4) and Google Chrome (Version 99.0.4844.84 (Official Build) (x86_64))
- Operating system: macOS Monterey (12.3)
- Ruby version:  ruby 2.6.8p205
- Bundler version:  Bundler version 1.17.2
- Jekyll version:  jekyll (4.2.2)

### Checklist

- [ ] I have performed a self-review of my code
- [ ] I have commented on my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings

Issue #543, opened 2022-03-31
Horizontal Scrolling for Long Equations


Jayitha:
## Checklist

- [x] I have read the [contributing guidelines](https://github.com/cotes2020/jekyll-theme-chirpy/blob/master/.github/CONTRIBUTING.md).
- [x] There is no similar request on [existing issues](https://github.com/cotes2020/jekyll-theme-chirpy/issues?q=is%3Aissue) (including closed ones).
- [x] I have read the [project progress](https://github.com/cotes2020/jekyll-theme-chirpy/projects) and know the current progress of the project.
- [x] I was in the `master` branch of the latest code.

## Is your feature request related to a problem? Please describe

Currently, long equations rendered by MathJax tend to overflow the post content onto the right side navigational toc. For instance, when I included the following equation

```
$$
(a + b)^3 = {3 \choose 0} a^3 b^0 + {3 \choose 1} a^2 b^1 + {3 \choose 2} a^1 b^2 + {3 \choose 0} a^0 b^3 + 0 \left ({3 \choose 0} a^3 b^0 + {3 \choose 1} a^2 b^1 + {3 \choose 2} a^1 b^2 + {3 \choose 0} a^0 b^3 \right )
$$
```

<img width="876" alt="image" src="https://user-images.githubusercontent.com/17466327/161130918-70e442e2-36e0-4d5b-95da-101a1ef06365.png">

This problem is very stark when viewing on a mobile, even relatively small equations tend to get cut off and the rest of the equation is not viewable. 

## Describe the solution you'd like

Ideally, I would like to have the ability to **scroll through the equation horizontally** when it exceeds the width of the post. Similar to how code blocks behave. 

## Describe alternatives you've considered

I've currently hacked together a solution to enable horizontal scrolling for equations by including the following piece of code[^source] in the `_sass/addon/commons.scss` file

```css
mjx-container {
  display: inline-grid;
  overflow-x: auto;
  overflow-y: hidden;
  max-width: 100%;
}
```
[^source]: [StackOverflow Answer](https://stackoverflow.com/a/65381823)

This is the way my hacky solution seems to work (also demo [here](https://jayitha.github.io/posts/text-and-typography/#mathematics))

https://user-images.githubusercontent.com/17466327/161134382-bf26fa7a-65b5-41fb-8966-a9d24c6b9027.mov

[...]

PR #2770, opened 2026-07-06
fix(math): prevent long display equations from overflowing the page

cornpip:
## Type of change

## Description

On mobile-width viewports, a display equation wider than the content column extends past the right edge of the page and causes **page-level horizontal scrolling** — the whole layout shifts when swiping, and after a portrait ⇄ landscape rotation the page can end up with leftover horizontal scroll or not filling the viewport width.

This is a regression of #543 / #545: the `mjx-container` overflow rules were dropped when MathJax was bumped to v4 in 97a537e. MathJax 4 has a native option for exactly this — [`output.displayOverflow`](https://docs.mathjax.org/en/latest/options/output/index.html) — but it defaults to `'overflow'` (content just spills out). This PR sets it to `'scroll'`, so a long display equation scrolls inside its own container instead of widening the page.

**How to reproduce**: create a post with `math: true` and a display equation wider than the viewport, e.g.

```latex
$$
\mathbf{T} = \arg\min_{\mathbf{R},s,\mathbf{t}} \sum_{i=1}^{N} w_i \left\| s\mathbf{R}\mathbf{a}_i + \mathbf{t} - \mathbf{b}_i \right\|^2, \quad \text{subject to } \mathbf{R}^\top\mathbf{R} = \mathbf{I}, \; \det(\mathbf{R}) = 1, \; s > 0
$$
```

Measured on a 390 px viewport (Chrome, Pixel 8 emulation): `document.documentElement.scrollWidth` is **733 px before** and **390 px after** this change; the equation scrolls within its own box and inline math is unaffected. Also verified against the `text-and-typography` demo post (including AMS-numbered equations) with no visual changes.

| Before (page overflows) | After (equation scrolls in place) |
| --- | --- |
| before | after |

*(right screenshot shows the same equation scrolled to its end inside the container)*

An alternative would be restoring the old CSS (`mjx-container { overflow-x: auto; overflow-y: hidden; }`), which also works, but `displayOverflow` is the v4-native mechanism and is renderer-agnostic, so I went with the config option.

## Additional context

Regression introduced in 97a537e (MathJax 3.2.2 → 4.x); originally fixed by #545 for #543.

### DIFF
--- a/assets/js/data/mathjax.js
+++ b/assets/js/data/mathjax.js
@@ -21,5 +21,9 @@ MathJax = {
     ],
     {%- comment -%} equation numbering {%- endcomment -%}
     tags: 'ams'
+  },
+  output: {
+    {%- comment -%} scroll long display equations instead of overflowing the page {%- endcomment -%}
+    displayOverflow: 'scroll'
   }
 };


cotes2020 (approved):
LGTM

PR #2771, opened 2026-07-06
fix(ui): truncate code block labels that overflow the viewport

cornpip:
## Type of change

## Description

The code block header label is rendered from `data-label-text` (`span::after`) without any width constraint. `.code-header` is a flex row and the label's default `min-width: auto` prevents the flex item from shrinking, so a long label — e.g. a long `file` path — pushes past the right edge of the viewport on mobile: the copy button is shoved off-screen and the page gets **page-level horizontal scrolling** (which also misbehaves across portrait ⇄ landscape rotations).

This PR lets the label flex item shrink (`min-width: 0`) and truncates it with an ellipsis, keeping the copy button in place.

**How to reproduce**: create a post with a code block labeled with a long file path:

````markdown
```cpp
int main() { return 0; }
```
{: file="mediapipe/modules/face_geometry/data/geometry_pipeline_metadata_landmarks.pbtxt" }
````

Measured on a 390 px viewport (Chrome, Pixel 8 emulation): `document.documentElement.scrollWidth` is **600 px before** and **390 px after** this change. Short labels (language names) are unaffected.

| Before (label overflows, copy button off-screen) | After (truncated, copy button visible) |
| --- | --- |
| before | after |

## Additional context

`npm run lint:scss` passes.

### DIFF
--- a/_sass/base/_syntax.scss
+++ b/_sass/base/_syntax.scss
@@ -186,7 +186,12 @@ div {
 
   /* the label block */
   span {
+    /* truncate long labels instead of overflowing the viewport */
+    overflow: hidden;
+    min-width: 0;
     line-height: v.$code-header-height;
+    text-overflow: ellipsis;
+    white-space: nowrap;
 
     /* label icon */
     i {
--- a/_sass/base/_syntax.scss
+++ b/_sass/base/_syntax.scss
@@ -188,7 +188,6 @@ div {
   span {
     /* truncate long labels instead of overflowing the viewport */
     overflow: hidden;
-    min-width: 0;
     line-height: v.$code-header-height;
     text-overflow: ellipsis;
     white-space: nowrap;
--- a/_sass/base/_syntax.scss
+++ b/_sass/base/_syntax.scss
@@ -132,7 +132,12 @@ div[class^='language-'] {
       &::before {
         content: '';
         display: inline-block;
+        flex-shrink: 0;
         margin-left: $dot-margin;
+
+        /* the 2nd/3rd dots are painted with box-shadow and take no layout
+          space — reserve it so a truncated label cannot overlap them */
+        margin-right: (v.$code-dot-size + v.$code-dot-gap) * 2;
         width: v.$code-dot-size;
         height: v.$code-dot-size;
         border-radius: 50%;
@@ -145,8 +150,9 @@ div[class^='language-'] {
       }
 
       span {
-        // center the text of label
-        margin-left: calc(($dot-margin + v.$code-dot-size) / 2 * -1);
+        /* fill the space between the dots and the copy button */
+        flex: 1;
+        text-align: center;
       }
     }
   }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: _sass/base/_syntax.scss [hunk 1]
```diff
@@ -186,7 +186,12 @@ div {
 
   /* the label block */
   span {
+    /* truncate long labels instead of overflowing the viewport */
+    overflow: hidden;
+    min-width: 0;
```
cotes2020:
```css
 min-width: 0;
 ```

Could you please tell me what this line of code does?

cornpip: Turns out this was redundant. Removed in 267d29d.

Per [CSS Flexbox §4.5](https://www.w3.org/TR/css-flexbox-1/#min-size-auto), `overflow: hidden` already sets the flex item's automatic minimum size to zero, so truncation works without this line. Verified in the latest Chrome, Firefox, and WebKit.

cotes2020 (changes requested):
There are still two obvious issues left unresolved:

1. The three virtual buttons on the left are blocked by the label text.

Image

2. The color of the label text's ellipsis does not match the text itself.

Image


---

https://github.com/user-attachments/assets/f098f8d9-612a-4ce6-bb30-94bb0e8eeaed

cornpip:
Both fixed in af7b712:

1. Dots overlap - the 2nd/3rd dots are `box-shadow` (no layout width), so I reserve their space on `::before` and let the label fill the gap with `flex: 1` instead of the negative `margin-left`.
2. Ellipsis color - the ellipsis is drawn by the `span`, not `::after`, so I moved `color` onto the `span`; `::after` inherits it now.

Short labels now center between the dots and the copy button, so they sit slightly further right than before.

cotes2020 (commented):
Just out of curiosity, are your PR description and all the code changes AI-generated?

cornpip:
Yeah, AI-assisted. I hit this using the theme on my own blog, so I used AI to investigate and write the fix. I went through each change myself.

cotes2020:
I appreciate your enthusiasm and willingness to contribute to the project, but I don't appreciate directly copy-pasting AI-generated content like this.

When it comes to PR descriptions or commit messages, simple issues can be described concisely, whereas AI tends to produce long blocks of text and citations, much like an academic report, which increases the time it takes for maintainers to read through them.

As for AI-generated or modified code, while it may solve the problem, it is not perfect. Often, it contains irrelevant or redundant code that not only lowers code quality but also increases future maintenance costs. If you don’t understand why the code is written a certain way, you won’t be able to identify and remove these minor flaws. Pasting it directly into a PR places an additional burden on the project maintainers.

Therefore, please review and digest the AI-generated content first, optimize it yourself, and then submit it. If this is beyond your technical expertise, consider submitting a bug report instead of a PR. :)

cornpip:
Got it. Thanks for taking the time to explain your standards. I appreciate it. 😊

